PR #2, opened 2018-01-22
explicit error message on unmapped pair

MarechJ:
Hi, while trying to use your lib I hit the following exception:

```
Traceback (most recent call last):
....
    return _std_trading_pairs[pair][exchange]
KeyError: 'BITFINEX'
```

Even though I was using only pairs pulled from:  https://api.bitfinex.com/v1/symbols"

So I modified the code a bit to figure out which pair was causing the problem. 

The pairs:
    #- [fun,usd]
    #- [fun,btc]
    #- [fun,eth]
    #- [zrx,usd]
    #- [zrx,btc]
    #- [zrx,eth]

Cheers

### DIFF
--- a/cryptofeed/standards.py
+++ b/cryptofeed/standards.py
@@ -2348,7 +2348,13 @@ _exchange_to_std = {
 
 def pair_std_to_exchange(pair, exchange):
     if pair in _std_trading_pairs:
-        return _std_trading_pairs[pair][exchange]
+        try:
+            return _std_trading_pairs[pair][exchange]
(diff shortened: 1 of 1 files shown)


bmoscon:
Ok, its not a bad check to have. FYI though, I support V2 of the API:

https://bitfinex.readme.io/v2/docs/ws-general

and the bitfinex trading pairs it supports are in bitfinex/pairs.py though those are the un-normalized (i.e. bitifinex specific) pairs.

Issue #4, opened 2018-01-22
GDAX trades channel and trade ids


MarechJ:
Hi,

I think the GDAX trades channel should be emulated from the 'ticker' one as it contains price, side and size for the last trade, which is pretty much what Bitfinex gives in their trades channels. It's not convenient to also have to add the oderbook. What I offer would be "aggregated trades" similar to what binance does. The current functionality could be separated as a "granular trades" (binance has that too).

Side note and less important: I think it's a shame to exclude ids from msgs (trade id for example), many implementations might need it (we do). Have a look at what the ccxt lib does, IMO it's a good level of abstraction and standardization but they also leave the raw message in the event for clients that need to specialize.

bmoscon:
both sound doable, will get to it soon - thanks for submitting an issue

chellowlolo:
For what’s its worth I agree completely with this suggestion.

MarechJ:
I'de be interesting to standardize the names of the channels, for examples if I start a feed on GDAX for 'trades' it listens to 'ticker' or 'match' behind the scene. I'll see if I have time for a PR.

bmoscon:
I've already done something like that for some feeds, but it will be consistent across all of them soon.

bmoscon:
I believe this is fixed now (I just added trade ids for gemini and gdax)

PR #7, opened 2018-01-25
GDAX channel name standardized, trades feed from ticker, book from level2 feed

MarechJ:
Following the comment in my open issue #4 , here's a candidate PR.

### DIFF
--- a/cryptofeed/gdax/gdax.py
+++ b/cryptofeed/gdax/gdax.py
@@ -17,11 +17,16 @@ from cryptofeed.callback import Callback
 class GDAX(Feed):
     def __init__(self, pairs=None, channels=None, callbacks=None):
         super(GDAX, self).__init__('wss://ws-feed.gdax.com')
-        self.channels = channels
+        self.user_channels = channels
+        # user ticker channel for trades and match for book
+        channels_map = {'trades': 'ticker', 'book': 'level2'}
+        self.channels = [channels_map.get(c, c) for c in channels]
         self.pairs = pairs
         self.book = {}
+        self.level2 = {}
         self.order_map = {}
         self.seq_no = {}
+        self.level2_book = {}
         self.callbacks = {'trades': Callback(None),
                           'ticker': Callback(None),
                           'book': Callback(None)}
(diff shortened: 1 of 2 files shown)


bmoscon:
I'm not sure why you'd remove the L3 book support. At some point I want to let users choose between L2 and L3 books - for feeds that dont offer L2 books, I'd provide my own (i.e. top 100 (or whatever) levels from L3 book. I'm fine adding L2 book support, but not at the expense of L3 books

MarechJ:
The L3 book should still work as before, unless I missed something? 
However I had trouble running it, when using all available pairs, the http calls in the _book_snapshot was hitting a rate limit (due to parallelization) i guess, and was crashing the whole thing. Also the format diverged from the bitfinex one (decimal keys instead of floats), and I wasn't sure if it's normal or not.

Please suggest edits, I'd rather contribute to this project rather than staying on my fork :) But I need a completely unified output format and aggregated book.

bmoscon:
I'd be fine taking this (and maybe tweaking some of the things) but I'm not understanding why the versions are pinned. I really dont like version pinning unless there is a reason. Does some part of this break or not work with lower versions?

MarechJ:
It's pinned to a minimum version updates will still go in (however that could cause trouble in the future). I did this because the websockets dependancy would not install for some reason, both in my docker setup and local virtualenv.

I'll start integrating you lib further in our project and might come back with similar PRs :)

How often do you push to pypi?

bmoscon:
Ok, I'll get this merged sometime today. I'll release a new version to pypi sometime this weekend. In general I've only been pushing to pypi when there was a reason to (new feed, bugfix, or new functionality), but I can push whenever :)

bmoscon:
sorry - another question. Why do you want the trades to come from the ticker? The matches feed is a true, real time source of ALL trades whereas the ticker is a batched channel so you wont really get a notification every time a match occurs. @MarechJ

bmoscon:
@MarechJ ok reading over your issue again, it seems this is really "aggregated trades", right? Just wanted to make sure I'm understanding the intent

MarechJ:
Yes, aggregated trades is what the trader I'm working with wants for some sophisticated quantitative analysis based on actual volumes. Ff the trades are aggretated from the taker point of view, it does not really change much and i'd say that 99% of the time you don't need to know all the small fills of a taker's side order. Maybe for keeping the lib friendly we should establish some standard definitions of the channels types (whether they are emulated or not), something like:
- ticker: best bid/ask, 24h volume (rolling), last price
- trades (Aggregated), buy/sell (taker's side), average price (VWAP if possible), trade size, timestamp 
- matches (non aggregated trades): same as above but with exact prix
- book: (aggregated) top 50-100 best bid/ask, price, size, nb of trades (optionnal), timestamp
- full_book: same as above but non aggregated, timestamp

For storage performance it'd also be preferable to have the 'pair' in the book callback other when need to iterate on all the pairs even though only one has been updated.

For good UX also have the raw_message passed in the callback for every channel.

And last remark, the current format of the book with price as keys is handy for updating it, but not for using it, i need to sort it before storing it or before using it for analysis this could be problematic for strategies relying on perfs.

Let me know what you think.
Sorry if there are a lot of typos I'm in a bit of a rush :D 

Cheers

bmoscon:
everything you've suggested, except raw message request, I've already started working on in a branch, `standardize`

MarechJ:
Ok, great. If you have a detailed plan you can assign a few tickets to me, I can tackle some stuff next week. Thx

MarechJ:
I made a mistake converting prices to float, you actually use Decimal in other exchanges.

PR #10, opened 2018-01-31
Revert back to decimal

MarechJ:
For some reason I thought the rest of the lib used floats, I reverted to Decimal for standardization's sake


bmoscon:
can you undo the version bump? Any time i release a version to pypi I bump the version (so the version in master is greater than that in pypi). So it shouldnt be bumped until another release is completed.

MarechJ:
Ok done. I'll wait til the other branch is stable before doing more PRs :)

PR #11, opened 2018-02-11
Update to standardize channels, feeds, exchanges, etc

bmoscon:
* major overhaul to remove need to type out strings for different things. Most things should have a defined constant you can import and use
* Book updates only return the book for the updated pair, not the entire book for the feed


bmoscon:
@MarechJ take a look. I've probably missed updating some part of this on a feed or two, but I believe its 95% here.

MarechJ:
Good, stuff! I'll have a look asap, probably tomorrow.

bmoscon:
@MarechJ going to merge this in. if you have any comments or things you want changed, you can still comment after the fact and I'll consider them and open new issues as appropriate. need to get this merged in so i can push out some other changes i have that are pending

bmoscon:
@MarechJ yes if possible most the channels (L2, ticker, trade, etc) should be emulated if not available. Thats the plan at least :)

PR #24, opened 2018-05-28
synthesized feed method, L3BookUpdates

rjbks:
Periodic calls using synthesize_feed only tested with GDAX currently


rjbks:
I actually pushed the wrong changes, I will close this and open another PR.

PR #26, opened 2018-06-29
Order book deltas

bmoscon:
* If a feed is subscribed with the `BookUpdateCallback, book deltas will be sent, with periodic full book updates.


bmoscon:
@rjbks take a look at this. This is more what I had in mind. I'm still open to your approach, but i feel like this is a lot easier to understand and the code is simpler. Thoughts?

PR #34, opened 2018-09-19
Gdax api

PandaXcentric:


cryptofeed/rest/gdax.py
```diff
@@ -0,0 +1,256 @@
...
+from cryptofeed.standards import pair_std_to_exchange, pair_exchange_to_std
+
+REQUEST_LIMIT = 10
+LOG = get_logger('rest', 'rest.log')
+
+# API Docs https://docs.gdax.com/
+class Gdax(API):
+    ID = GDAX
+    # api = "https://api.gdax.com"
+    api = "https://api-public.sandbox.gdax.com"
```
bmoscon: maybe have a way to select sandbox vs real api? Default should be real, but maybe change API's `__init__` to allow a sandbox=True kwarg

cryptofeed/rest/gdax.py
```diff
@@ -0,0 +1,256 @@
...
+
+    
+    def _generate_signature(self, endpoint: str, method: str, body = ''):
+        timestamp = str(time.time())
+        message = ''.join([timestamp, method, endpoint, (body or '')])
```
bmoscon: `body` defaults to empty str, so you just need `body` there

PR #35, opened 2018-09-22
Gdax api

PandaXcentric:


cryptofeed/rest/gdax.py
```diff
@@ -35,6 +35,21 @@ def _generate_signature(self, endpoint: str, method: str, body = ''):
             'Content-Type': 'Application/JSON',
         }
 
+    def _pagination(self, endpoint: str, body={}):
```
bmoscon: the other part of gdax uses '' as the empty/default body. Why not here?

cryptofeed/rest/gdax.py
```diff
@@ -174,8 +191,16 @@ def _delete_order(self, order_id=None):
         if order_id is not None:
             endpoint = '{}/{}'.format(endpoint, order_id)
 
-        header = self._generate_signature(endpoint, "DELETE", body=json.dumps(body))
+        header = self._generate_signature(endpoint, "DELETE")
```
bmoscon: you're creating a var `header` but below using `headers` in the method call. I think this is a bug - guessing this one isn't testable on the sandbox?

cryptofeed/rest/gdax.py
```diff
@@ -234,9 +259,90 @@ def get_orders(self, body):
...
+        data format
+        {
+            "amount": 10.00,
+            "currency": "USD",
+            "payment_method_id": "bc677162-d934-5f1a-968c-a496b1c1270b"
+        }
+        """
+        return self._post("/deposits/payment-method", body)
+
+    def deposit_coinbase(self, body):
```
bmoscon: the other one is called deposit funds, why call this one "coinbase"? Why not `deposit_coins` ?

cryptofeed/rest/gdax.py
```diff
@@ -234,9 +259,90 @@ def get_orders(self, body):
...
+        data format
+        {
+            "amount": 10.00,
+            "currency": "USD",
+            "payment_method_id": "bc677162-d934-5f1a-968c-a496b1c1270b"
+        }
+        """
+        return self._post("/withdrawals/payment-method", body)
+    
+    def withdrawl_coinbase(self, body):
```
bmoscon: you spelled withdrawals wrong ;)

cryptofeed/rest/gdax.py
```diff
@@ -234,9 +259,90 @@ def get_orders(self, body):
...
+        data format
+        {
+            "amount": 10.00,
+            "currency": "BTC",
+            "coinbase_account_id": "c13cd0fc-72ca-55e9-843b-b84ef628c198",
+        }
+        """
+        return self._post("/withdrawals/coinbase-account", body)
+    
+    def withdrawl_crypto(self, body):
```
bmoscon: same spelling error

bmoscon: also, maybe use coin here?

cryptofeed/rest/gdax.py
PandaXcentric: it's '' in one place because it needs to be string, it's None in the other place body is supposed to be dict, so defaulting it to that

cryptofeed/rest/gdax.py
PandaXcentric: because this is specifically letting you deposit from a different coinbase account, if they integrate with other crypto accounts then this will be how we distinguish...funds is just from any other payment method

bmoscon: true, but then you'd end up with deposit_bitfinex, deposit_bitmex, deposit_coinbase, etc and they should all have standardized names. you already distinguish the exchange by the attr on the rest class, i dont think we need it on the method name too, do we?

PR #42, opened 2018-10-14
Fixing bug in coinbase level 2 update logic

ejp314:
I was running demo_nbbo.py for Coinbase only and realized it would stop updating if the size at a given price level got to 0. The logic is there to delete the level in the sorted dictionary if the amount goes to 0, but it never gets into the if-statement because of a type incompatibility.

I'm not sure if this issue exists for other exchanges, but this appears to fix it for Coinbase.

Great piece of code, by the way. Thanks for sharing!


bmoscon:
Ah you're right - the fix is a bit different though - it just needs to be `0`, not `"0"`. If you change it, I'll accept it:


```
In [9]: 0 == decimal.Decimal(0.0) == decimal.Decimal(0)                                                                                                      
Out[9]: True
```

bmoscon:
@ejp314 ^ see above

ejp314:
Good catch - I think it's fixed now. Feel free to accept or make the change on your own if that's easier!

PR #44, opened 2018-10-20
Kraken api

PandaXcentric:


cryptofeed/rest/kraken.py
```diff
@@ -0,0 +1,279 @@
...
+        if resp.status_code != 200:
+            LOG.error("%s: Status code %d", self.ID, resp.status_code)
+            LOG.error("%s: Headers: %s", self.ID, resp.headers)
+            LOG.error("%s: Resp: %s", self.ID, resp.text)
+            resp.raise_for_status()
+
+        return resp.json()
+
+
+    def _post_private(self, command: str, payload = {}):
```
bmoscon: setting default args to empty dicts are bad news. Set them to None and set to {} inside of the func if necessary

cryptofeed/rest/kraken.py
```diff
@@ -0,0 +1,279 @@
+from time import time
+import hashlib, hmac, requests, urllib, json, base64
+
+from cryptofeed.rest.api import API
+from cryptofeed.feeds import KRAKEN
```
bmoscon: you should import it from cryptofeed.exchanges since thats the place they are all defined

PR #45, opened 2018-10-21
Fixing issue with bitstamp order book message parsing

ejp314:
I think we need to convert price and size strings in order book messages to Decimal objects


bmoscon:
ok let me quickly confirm the json call isnt doing the conversion and then I'll merge this

PR #52, opened 2019-01-09
#51 Fix to provide a correct pair in Binance book feed

### LINKED ISSUE #51: Fix to provide a correct pair in Binance book feed
Now the order book from binance feed may indicate the full subscription list rather than the updated instrument due to #50 

```
await self.callbacks[L2_BOOK](feed=self.id, pair=self.pairs, book=self.l2_book, timestamp=time.time() * 1000)
```

According to binance websocket [documentation](https://github.com/binance-exchange/binance-official-api-docs/blob/master/web-socket-streams.md), the instrument name is provided in the "stream"

gavincyi:
The change is to resolve #51 bug

### DIFF
--- a/cryptofeed/binance/binance.py
+++ b/cryptofeed/binance/binance.py
@@ -144,6 +144,8 @@ class Binance(Feed):
             await self._trade(msg)
         elif msg['e'] == '24hrTicker':
             await self._ticker(msg)
+        else:
+            LOG.warning("%s: Unexpected message received: %s", self.id, msg)
 
     async def subscribe(self, websocket):
         pass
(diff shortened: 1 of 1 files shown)


cryptofeed/binance/binance.py
```diff
@@ -126,21 +126,24 @@ def __reset(self):
...
+        pair = pair_exchange_to_std(pair.upper())
+
+        if event == 'depth20':
+            await self._book(msg, pair)
         elif msg['e'] == 'trade':
             await self._trade(msg)
         elif msg['e'] == '24hrTicker':
             await self._ticker(msg)
-        else:
-            LOG.warning("%s: Unexpected message received: %s", self.id, msg)
```
bmoscon: why remove the log warning? Does the change to streams mean we'll get unhandled messages?

Issue #60, opened 2019-02-03
SSL issue after updating to current master (both win 10 and mac)


ypetrusha:
Launching any of examples I end up with the following either on mac or win10:

> ...   
  File "C:\Python37\lib\ssl.py", line 412, in wrap_socket
    session=session
  File "C:\Python37\lib\ssl.py", line 853, in _create
    self.do_handshake()
  File "C:\Python37\lib\ssl.py", line 1117, in do_handshake
    self._sslobj.do_handshake()
TimeoutError: [WinError 10060] A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond

bmoscon:
That sounds like a problem with either your environment or your connection, especially since the errors are coming from the python internals

bmoscon:
I did just push an update to master, but I dont think its related to your issue.

bmoscon:
I'd be happy to try on my end too if you can provide me with a small script that reproduces the issue. I was able to run `examples/demo.py`, among others, so I'm pretty confident the issue is not in this library

PR #64, opened 2019-02-04
#60 Fix system crash if Bitstamp not available

petrury:
This PR to address the problem when app crashes when Bitstamp is not available (e.g. banned by provider).
Please review. Probably it's not good idea to patch for Bitstamp only.

### DIFF
--- a/cryptofeed/bitstamp/pairs.py
+++ b/cryptofeed/bitstamp/pairs.py
@@ -5,7 +5,9 @@ Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
 import requests
+import logging
 
+LOG = logging.getLogger('bitstamp.pairs')
 
 def gen_pairs():
     ret = {}
@@ -16,5 +18,8 @@ def gen_pairs():
         ret[normalized] = pair
     return ret
 
-
-bitstamp_pair_mapping = gen_pairs()
+try:
+    bitstamp_pair_mapping = gen_pairs()
+except:
+    LOG.exception("Unable to get pairs from Bitstamp")
+    bitstamp_pair_mapping = None
--- a/cryptofeed/standards.py
+++ b/cryptofeed/standards.py
@@ -45,12 +45,15 @@ mappings = {
 }
 
 for exchange, mapping in mappings.items():
-    for std, exch in mapping.items():
-        _exchange_to_std[exch] = std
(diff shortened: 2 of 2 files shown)


bmoscon:
@ypetrusha I can merge this as-is and make the changes to it I want, or I can provide comments about what I'd want changed (there would be a lot of changes). Up to you, let me know!

ypetrusha:
@bmoscon I think I can try. Please, provide the details.

cryptofeed/bitstamp/pairs.py
```diff
@@ -5,7 +5,9 @@
 associated with this software.
 '''
 import requests
+import logging
 
+LOG = logging.getLogger('bitstamp.pairs')
```
bmoscon: use the same file the others use, see any of the exchange classes

cryptofeed/bitstamp/pairs.py
```diff
@@ -16,5 +18,8 @@ def gen_pairs():
         ret[normalized] = pair
     return ret
 
-
-bitstamp_pair_mapping = gen_pairs()
+try:
+    bitstamp_pair_mapping = gen_pairs()
+except:
+    LOG.exception("Unable to get pairs from Bitstamp")
+    bitstamp_pair_mapping = None
```
bmoscon: dont set to none, set to a hard coded list of pairs, see https://github.com/bmoscon/cryptofeed/blob/master/cryptofeed/gemini/pairs.py as an example

cryptofeed/standards.py
```diff
@@ -45,12 +45,15 @@
 }
 
 for exchange, mapping in mappings.items():
-    for std, exch in mapping.items():
-        _exchange_to_std[exch] = std
-        if std in _std_trading_pairs:
-            _std_trading_pairs[std][exchange] = exch
-        else:
-            _std_trading_pairs[std] = {exchange: exch}
+    if mapping:
```
bmoscon: not necessary with change mentioned above

bmoscon:
@ypetrusha i actually changed how the pairs work - they're only generated now for configured exchanges, which i think is a better approach, so this wont be necessary

ypetrusha:
Thank you @bmoscon 
Now it works as is for me. I'm closing this this PR.

PR #66, opened 2019-02-09
In trades replace BID / ASK with BUY / SELL.

ooinaru:
Primarily semantics changes, but I feel like trades side should either be BUY or SELL instead of BID / ASK, which is used when describing the order book. 
Also, side on Binance can be had from side if buyer is maker.

### DIFF
--- a/cryptofeed/binance/binance.py
+++ b/cryptofeed/binance/binance.py
@@ -12,7 +12,7 @@ import time
 from sortedcontainers import SortedDict as sd
 
 from cryptofeed.feed import Feed
-from cryptofeed.defines import TICKER, TRADES, BUY, SELL, BID, ASK, L2_BOOK, UND, BINANCE
+from cryptofeed.defines import TICKER, TRADES, BUY, SELL, BID, ASK, L2_BOOK, BINANCE
 from cryptofeed.standards import pair_exchange_to_std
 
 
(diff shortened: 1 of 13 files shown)


cryptofeed/binance/binance.py
```diff
@@ -60,7 +60,7 @@ def __reset(self):
         await self.callbacks[TRADES](feed=self.id,
                                      order_id=msg['t'],
                                      pair=pair_exchange_to_std(msg['s']),
-                                     side=UND,
```
bmoscon: if you are no longer using UND you should remove it from the imports

bmoscon:
also please update the CHANGES file

cryptofeed/gemini/gemini.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, pairs=None, channels=None, callbacks=None, **kwargs):
 
     async def _trade(self, msg, timestamp):
         price = Decimal(msg['price'])
-        side = BID if msg['makerSide'] == 'bid' else ASK
+        side = BUY if msg['makerSide'] == 'bid' else SELL
```
ooinaru: @bmoscon Can you chack this works correctly? I don't have access to Gemini where I am from but it looks to me as if this should be `side = BUY if msg['makerSide'] == 'ask' else SELL`,

bmoscon: Looks right and matches the docs. I'll give it a test today

PR #75, opened 2019-04-09
Handle binance None pair

Werdor:
Fix exception raised on binance when a symbol specified in channel is not found.  

 File "/home/werdor/.local/lib/python3.7/site-packages/cryptofeed/binance/binance.py", line 27, in __init__
    self.address = self.__address()
  File "/home/werdor/.local/lib/python3.7/site-packages/cryptofeed/binance/binance.py", line 34, in __address
    pair = pair.lower()
NoneType: None
2019-04-10 00:09:26 ERROR  root                 start.py:49       <class 'AttributeError'>: 'NoneType' object has no attribute 'lower'
NoneType: None

### DIFF
--- a/cryptofeed/binance/binance.py
+++ b/cryptofeed/binance/binance.py
@@ -31,9 +31,10 @@ class Binance(Feed):
         address = "wss://stream.binance.com:9443/stream?streams="
         for chan in self.channels if not self.config else self.config:
             for pair in self.pairs if not self.config else self.config[chan]:
-                pair = pair.lower()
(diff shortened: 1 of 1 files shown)


bmoscon:
I think supplying an invalid trading pair and not getting an error is a bad idea. I'm fine with changing the somewhat cryptic error (NoneType) into an exception. You could raise the exception where the normalized trading pair gets converted into the exchange formatted trading pair. Probably would want to define a new exception for this.

bmoscon:
@Werdor think you'll make that change? if not, I can

PR #80, opened 2019-04-20
Rewrite book delta demo using DeltaBook class.

clefru:
This rewrites the demo_book_delta.py example to a Class to handle multiple feed checkings at the same time. I do think that there is value in promoting DeltaBook to a more generic place in the codebase, potentially in cryptofeed/callbacks.py, but that would require giving the interface a bit more thought, in particular what to do when the book equality check fails. I'll send a follow-up PR to discuss this, after I played around with some variants.

(Also I do think that the check_books function is incomplete in checking book equality)

### DIFF
--- a/examples/demo_book_delta.py
+++ b/examples/demo_book_delta.py
@@ -12,77 +12,76 @@ from cryptofeed.exchanges import Bitmex, Coinbase, Bitfinex, Gemini, HitBTC, Pol
 from cryptofeed.defines import L2_BOOK, L3_BOOK, BID, ASK, BOOK_DELTA
 
 
-BOOK = None
+class DeltaBook(object):
+    def __init__(self, name):
+        self.book = None
+        self.name = name
+        self.L2 = {L2_BOOK: BookCallback(self.handle_book),
+                   BOOK_DELTA: BookUpdateCallback(self.handle_l2_delta)}
+        self.L3 = {L3_BOOK: BookCallback(self.handle_book),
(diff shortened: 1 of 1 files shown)


bmoscon:
thanks for the PR!

I'm not sure what you mean by making book delta more generic - I feel like its pretty generic already. An alternative could be to only have a "book update" callback that has an arg in it that indicates the update type (full book or delta). Is that what you had in mind? Something else?

clefru:
@bmoscon: DeltaBook uses incoming deltas to build a synthetic book from book & delta callbacks (in particular the l2_delta/l3_delta handlers that maintain self.book[side][price]([order]). I feel this is needed in most exchange code. For instance,
https://github.com/bmoscon/cryptofeed/blob/master/cryptofeed/bitfinex/bitfinex.py#L186
is quite identical to:
https://github.com/bmoscon/cryptofeed/blob/master/examples/demo_book_delta.py#L67
I feel that this should be unified.

bmoscon:
oh i see, i misunderstood your comment. You want a generic object for handling book deltas - one that would be used internally and be available externally if so desired. Is this correct? I'm not sure how useful it would be to any external code - you can just as easily get the full book on every update, its already being stored internally. The main advantage to the deltas (to me at least) is for storage - if you're storing the book to a database its preferable to store the deltas as opposed to the entire book. Definitely happy to discuss via email, or you can submit a PR with what you had in mind

PR #83, opened 2019-04-21
Use Bitstamp's V2 websocket API

edandavi:
V1 (pusher) will be discontinued in q2 2019.
Also v2 supports l3 oreder book.


bmoscon:
thanks for the PR!

PR #85, opened 2019-04-22
Validate Bitfinex funding channel in Bitfinex feed derived class.

edandavi:
I think that specific validation per exchange should be in the derived exchange object from feed.

P.S Kudos on a well designed project :)


bmoscon:
agree, its clearer this way

PR #98, opened 2019-05-25
Added Deribit Exchange

Vivikar:
Added Deribit support for TRADE and TICKER channels and BTC Perpetual pair.


cryptofeed/pairs.py
```diff
@@ -160,6 +160,9 @@ def coinbene_pairs():
     r = requests.get('http://api.coinbene.com/v1/market/symbol').json()
     return {f"{e['baseAsset']}-{e['quoteAsset']}" : e['ticker'] for e in r['symbol']}
 
+def deribit_pairs():
+    return 'BTC-PERPETUAL'
```
bmoscon: these should be dynamically generated, like the others. deribit has a rest endpoint where the specify the supported currencies (and they support more than just BTC)

cryptofeed/deribit/deribit.py
```diff
@@ -0,0 +1,144 @@
+import logging
+import json
+from decimal import Decimal
+import zlib
```
bmoscon: this import is not used

cryptofeed/deribit/deribit.py
```diff
@@ -0,0 +1,144 @@
...
+                        "params": {
+                            "channels": [
+                                f"{chan}.{pair}.raw"
+                            ]
+                        }
+                    }
+                ))
+
+    async def message_handler(self, msg):
+        # Huobi sends a ping evert 5 seconds and will disconnect us if we do not respond to it
```
bmoscon: this comment seems wrong . . .

bmoscon:
Looks a lot better @Vivikar - thanks! One last thing - can you rebase your fork? Its showing a number of changes in the diff that are ones I made after you forked. Lets clean that up so the history is correct

Vivikar:
> Looks a lot better @Vivikar - thanks! One last thing - can you rebase your fork? Its showing a number of changes in the diff that are ones I made after you forked. Lets clean that up so the history is correct

Done, I suppose

cryptofeed/deribit/deribit.py
```diff
@@ -0,0 +1,205 @@
...
+from decimal import Decimal
+
+LOG = logging.getLogger('feedhandler')
+
+
+class Deribit(Feed):
+    id = DERIBIT
+
+    def __init__(self, pairs=None, channels=None, callbacks=None, config=None, **kwargs):
+        super().__init__('wss://test.deribit.com/ws/api/v2', pairs=pairs,
```
bmoscon: thats the testnet endpoint right? not the live endpoint?

Vivikar: Yes, that's the test endpoint

bmoscon: @vivikar can we put in the correct endpoint for live data?

Vivikar: We can't. There is no live endpoint for Deribit api v2. As far as I understand api v2 is still under development, so they have only test endpoint. However the data we get is coming live, the only difference is that we get it through their test api.

bmoscon: i tested your code with wss://www.deribit.com/ws/api/v2 and it works, so I think it does exist, it may not be publicized though

cryptofeed/deribit/deribit.py
```diff
@@ -0,0 +1,205 @@
...
+                bidask[price] = amount
+                delta[ASK].append((Decimal(price), Decimal(amount)))
+            else:
+                del bidask[price]
+                delta[ASK].append((Decimal(price), Decimal(amount)))
+        await self.book_callback(msg["params"]["data"]["instrument_name"], L2_BOOK, False, delta, timestamp)
+
+    async def message_handler(self, msg):
+        # Converting msg we got from str to dict using yaml, because it contains singular quotes ' '
+        msg_dict = yaml.safe_load(msg)
```
bmoscon: dont use yaml, just use json.loads here. json parsing is significantly faster than yaml, even if you were using the C loader (which we are not). I tested it, and it works just fine with `json.loads`

PR #100, opened 2019-06-07
Fix L3 book update conversion

andimx:
This fixes a bug that occurs when using the L3_BOOK channel with a backend that supports delta updates.

Example code to reproduce:
```
from cryptofeed.exchanges import Coinbase
from cryptofeed.defines import L3_BOOK, BOOK_DELTA
from cryptofeed.backends.redis import BookRedis, BookUpdateRedis


def main():
    f = FeedHandler()
    f.add_feed(Coinbase(pairs=['BTC-USD'], channels=[L3_BOOK], callbacks={L3_BOOK: BookRedis(), BOOK_DELTA: BookUpdateRedis()}))
    f.run()


if __name__ == '__main__':
    main()
```

### DIFF
--- a/cryptofeed/backends/_util.py
+++ b/cryptofeed/backends/_util.py
@@ -17,10 +17,10 @@ def book_delta_convert(delta: dict, data: dict):
             else:
                 order_id, price, amount = entry
                 price = str(price)
-                if price in delta[side]:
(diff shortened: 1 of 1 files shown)


bmoscon:
good catch - thanks!

PR #112, opened 2019-07-06
Elastic Cache support

bmoscon:


cryptofeed/backends/_util.py
```diff
@@ -53,3 +53,27 @@ def book_convert(book: dict, data: dict, depth: int, convert=str):
         count += 1
         if depth and count >= depth:
             break
+
+
+def book_flatten(feed: str, pair: str, book: dict, timestamp: float, delta: str) -> dict:
```
PandaXcentric: it might be worth having 2 functions here, book_flatten_l2 and book_flatten_l3. They can be wrappers for the same shared code, but since L3 has the order_id field and L2 doesn't, it might be worth making them separate.

cryptofeed/backends/elastic.py
```diff
@@ -0,0 +1,102 @@
...
+        self.addr = f"{addr}/{index}/{index}"
+        self.session = None
+        self.numeric_type = numeric_type
+
+
+class TradeElastic(ElasticCallback):
+    def __init__(self, *args, index='trades', **kwargs):
+        super().__init__(*args, index=index, **kwargs)
+
+    async def __call__(self, *, feed: str, pair: str, side: str, amount: Decimal, price: Decimal, order_id=None, timestamp=None):
```
PandaXcentric: for this, it might be worth just letting it take the dict rather than individual fields, since the individual fields are the parameters that'll be in the trade dict anyway?

basically, for the use of this, do I need to break up a dict that has these same fields (if not then it's good as is).

cryptofeed/backends/elastic.py
```diff
@@ -0,0 +1,102 @@
...
+            'id': order_id,
+            'amount': self.numeric_type(amount),
+            'price': self.numeric_type(price),
+            'timestamp': timestamp
+        }
+        await self.write('POST', json.dumps(trade), headers={'content-type': 'application/json'})
+
+
+class FundingElastic(ElasticCallback):
+    def __init__(self, *args, index='funding', **kwargs):
```
PandaXcentric: whats the reason for *args rather than just taking addr (which seems like the only thing we'd expect there since you're passing it to the super which only takes that argument? This comment applies for all the classes that inherit from ElasticCallback

PandaXcentric: is it just so you can easily change the super class without having to modify all of the inheriting ones? I guess the tradeoff there would be readability of the inheriting class vs ease of modifying the super

cryptofeed/backends/elastic.py
```diff
@@ -0,0 +1,102 @@
...
+        super().__init__(*args, index=index, **kwargs)
+
+    async def __call__(self, *, feed, pair, **kwargs):
```
PandaXcentric: feed and pair aren't used here, is there a reason this function should still take them?

cryptofeed/backends/_util.py
bmoscon: true, but most the code is the same, so its a lot easier to have one extra check in here as opposed to duplicating the majority of the function.

cryptofeed/backends/elastic.py
bmoscon: you can still use the ** to turn a dict into kwargs

cryptofeed/backends/elastic.py
bmoscon: all funding callbacks have the same signature. it could be part of the update, but since elastic only lets you have a single type in an index, the index/type tell you the feed-pair combination so its not really necessary

PR #126, opened 2019-07-30
Update setup.py

xiandong79:
add "zmq" requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,7 +53,8 @@ setup(
         "pyyaml",
         "aiohttp",
         "aiodns",
-        "cchardet"
+        "cchardet",
+        "zmq"
     ],
     extras_require={
         'redis': ['aioredis'],


bmoscon:
its an optional requirement, why does it need to be required? 

https://github.com/bmoscon/cryptofeed/blob/master/setup.py#L61

Its only needed if one wants to make use of the ZMQ functionality the backend provides

xiandong79:
step1. I build the package of cryptostore + cryptofeed from source code in your Github.
step2. I run 'cryptostore --config my_config.yaml'. (I using Redis + influx + parquet backends)

I got 'ImportError: No module named zmq'. So it seems that it is a required package.

bmoscon:
zmq is only needed if you have passthrough enabled in the config. If its in there, you'd need ZMQ. You can install it with cryptostore[zmq] which will install the extras

xiandong79:
I do not need zmq as I am using Redis + influx + parquet as backends.

Do you mean that, I should comment following line in the config.yaml?
```
pass_through:
    type: zmq
    host: '127.0.0.1'
    port: 5678
```

xiandong79:
I always do not know what is the meaning of 'pass_through'.

I just leave it alone as the default config.yaml does. I notice that I do not use elastic for storage, but I can just leave the config of elastic alone.

PR #132, opened 2019-08-03
Added FTX.com feed

vdbelt:
Added support for spot and perpetual markets on FTX.com

https://docs.ftx.com/#websocket-api


bmoscon:
Awesome - thanks for the PR!

PR #150, opened 2019-08-29
Configurable binance book depth

delta-upsilon-pi:
Rate limiter is taking into account the depth `limit` parameter so makes sense to make it configurable.
Faced the case when _slow running_ instance which tries to handle many symbols easily banned provider's network.


bmoscon:
You'd need to do it the same way other feeds do this. See Kraken: https://github.com/bmoscon/cryptofeed/blob/9bd7c15a502fb196ff43196453c6c69c9857c595/cryptofeed/exchange/kraken.py#L24

Happy to do it if you dont have the time or want to, just let me know

delta-upsilon-pi:
https://github.com/bmoscon/cryptofeed/pull/151

PR #157, opened 2019-09-13
Renko Aggregation and using defines for strings

quantfiction:
Added RenkoFixed class to aggregate
Used defines rather than strings for feed id in Redis


bmoscon:
@quantfiction awesome! Any chance you can rebase or in some otherway fix the fork so it doesnt show 30+ commits coming in? I'd rather not make a mess of the commit history if possible - thanks!

PR #160, opened 2019-09-17
Develop

quantfiction:


xiandong79:
@quantfiction 

hi, what is the meaning of 'renko'? and why we/quant need it?

quantfiction:
Renko bars are a different way of aggregating candles, similar to the typical ohlcv [https://www.thebalance.com/technical-analysis-renko-chart-basics-1031113]

Added for use in a trading system that uses this bar type for signal generation.

PR #161, opened 2019-09-18
Add open interest functionality

quantfiction:
Created OPEN_INTEREST define, OpenInterestArctic callback, and example
file.

### DIFF
--- a/cryptofeed/backends/arctic.py
+++ b/cryptofeed/backends/arctic.py
@@ -85,12 +85,14 @@ class OpenInterestArctic(ArcticCallback):
         if self.key is None:
             self.key = OPEN_INTEREST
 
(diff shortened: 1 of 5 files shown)


bmoscon:
This is part of the instrument channel right? Why not just call the callback "instrument"? Open interest is a specific to futures/options exchanges, right? Do the others support this as well?

quantfiction:
I guess I'll have to explore how the other exchanges handle it. I'm not as interested in logging all of the information from each instrument update, and not all messages from the instrument channel have the same fields. In Bitmex's case, there's a lot of redundant information already captured in other channels

quantfiction:
Incomplete; need to complete backend for kafka

PR #174, opened 2019-11-02
Added feed for BitMax

vdbelt:


cryptofeed/exchange/bitmax.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class Bitmax(Feed):
+    id = BITMAX
+
+    def __init__(self, pairs=None, channels=None, callbacks=None, config=None, **kwargs):
+        if pairs is None:
+            pairs = list(set([pair for s in config.values() for pair in s]))
+        if len(pairs) != 1:
+            LOG.error("Bitmax requires a websocket per trading pair")
```
bmoscon: There are ways around this - it was previously done for gemini, see https://github.com/bmoscon/cryptofeed/blob/ef4b6621efa974be829380d22323e2c71696a7c8/cryptofeed/feedhandler.py#L216

vdbelt: thanks! I reintroduced the multi-pair method as you used with Gemini before, so it should be good now. 

If another exchange/feed comes along with the same requirements, i'll look into abstracting it away somehow.

PR #176, opened 2019-11-17
Huobi Support Ticker Data via "BBO" tag

aviatorBeijing:


vdbelt:
Since bbo effectively is a book update (L2_BOOK with depth 1), perhaps a different approach could be to subscribe to the bbo channel if max_depth=1 (using BookCallback). No callback signature changes required in that case.

bmoscon:
I'm inclined to agree with @vdbelt - if size is something you need, use the book data. Few (if any) of the other tickers have sizes, I'm not inclined to add it just for this one exchange. If I'm wrong and others do, let me know and I'll re-evaluate.

aviatorBeijing:
> I'm inclined to agree with @vdbelt - if size is something you need, use the book data. Few (if any) of the other tickers have sizes, I'm not inclined to add it just for this one exchange. If I'm wrong and others do, let me know and I'll re-evaluate.

Out of curiosity, what was the argument of excluding size from tick data? Compared to the book data, tick data is faster and data size is smaller than book data. Adding the size info in tick data could be helpful in many scenario.

aviatorBeijing:
> Since bbo effectively is a book update (L2_BOOK with depth 1), perhaps a different approach could be to subscribe to the bbo channel if max_depth=1 (using BookCallback). No callback signature changes required in that case.

Agree! Thanks.

Jay54520:
How about add a new ticker callback with sizes to make user has choice?
```
class TickerSizeCallback(Callback):
    async def __call__(self, *, feed: str, symbol: str, bid: Decimal, ask: Decimal, timestamp: float,
                       receipt_timestamp: float, bid_size: Decimal, ask_size: Decimal):
        await super().__call__(feed, symbol, bid, ask, timestamp, receipt_timestamp, bid_size, ask_size)
```

PR #191, opened 2019-12-17
Bug Fix & Cleanup

rshtirmer:


cryptofeed/backends/redis.py
```diff
@@ -88,9 +87,9 @@ class TickerStream(RedisStreamCallback, BackendTickerCallback):
     
 
 class OpenInterestRedis(RedisZSetCallback, BackendOpenInterestCallback):
-    default_key = OPEN_INTEREST
+    default_key = 'open_interest'
```
bmoscon: why hard code it instead of using the define? @rshtirmer

rshtirmer: Ah, I hard-coded it here just because all the others in redis.py where too. Wasn't sure if there was a reason for that, but figured I'd match it up with the rest of the file.

PR #193, opened 2020-01-04
Fix book backend missing feed and pair fields.

zhenpingfeng:
I hope this didn't break others backend.

### DIFF
--- a/cryptofeed/backends/backend.py
+++ b/cryptofeed/backends/backend.py
@@ -13,14 +13,14 @@ from cryptofeed.backends._util import book_convert, book_delta_convert
 
 class BackendBookCallback:
     async def __call__(self, *, feed, pair, book, timestamp):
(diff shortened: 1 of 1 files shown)


cryptofeed/backends/backend.py
```diff
@@ -13,14 +13,14 @@
 
 class BackendBookCallback:
     async def __call__(self, *, feed, pair, book, timestamp):
-        data = {'timestamp': timestamp, 'delta': False, BID: {}, ASK: {}}
+        data = {'feed': feed, 'pair': pair, 'timestamp': timestamp, 'delta': False, BID: {}, ASK: {}}
```
bmoscon: the data is already in the write below. If you need the data for rabbitmq, can't you just tweak its writes ?

zhenpingfeng: So, why the book data fields didn't include feed and pair, but trades, tick, and funding have? I think this breaks consistency. But I will try to tweak it.

Issue #194, opened 2020-01-05
Adding authentication token into header option for InfluxDB 2.0


heartcored98:
Since InfluxDB 2.0 requires token string within their request header, 

It would be great to pass the authentication token argument to the InfluxDB backend!

heartcored98:
I made a pull request #195 which solve this issue. Please take a look!

bmoscon:
merged

PR #195, opened 2020-01-05
Support InfluxDB 2.0 by adding token authentication headers (Issue #194)

heartcored98:
I am trying to construct my own crypto database, I choose to use InfluxDB 2.0 for several reasons. In order to solve #194, I made a few changes to your awesome code. 

### 1) Four additional parameter for InfluxCallback  
Since InfluxDB 2.0 always requires an authentication token when using the rest API method, I added four additional input arguments to `InfluxCallback`.

- org: str 
        Organization name for authentication  

- bucket: str  
Bucket name for authentication  

- token: str 
          Token string for authentication  

- precision: str 
          Precision level among (s, ms, us, ns)   

### 2) Change with `self.addr`.
Also, `self.addr` could be constructed in two ways.  
If the user provides all `org`, `bucket`, `token` arguments, then the address should be like below 
```
self.addr = f"{addr}/api/v2/write?org={org}&bucket={bucket}&precision={precision}"
``` 

Otherwise, the original style.  
```
self.addr = f"{addr}/write?db={db}"
```
 
### 3) Newly added `self.headers` containing authentication token  

I also created `self.headers` object in order to pass the authentication token.
This header sends to the http request method.   
```
self.headers = {"Authorization": f"Token {token}"}
```

### 4) Add brief demo with `demo_inflxudb.py` 
I also demonstrated how to deal with InfluxDB 2.0 by adding supplementary example code in the `demo_influxdb.py`  

### 5) Testing  
I tested the mentioned demo code and observed works well with InfluxDB Cloud 2.0 service. However, sometimes the Influx refuses connection due to a high request rate (so I need to work with cryptostore library). 

For summary, I hope you merge my pull request!

### DIFF
--- a/cryptofeed/backends/influxdb.py
+++ b/cryptofeed/backends/influxdb.py
@@ -63,7 +63,7 @@ class InfluxCallback(HTTPCallback):
           Precision level among (s, ms, us, ns) 
         """
         super().__init__(addr, **kwargs)
-        if (org and bucket and token):
+        if org and bucket and token:
             self.addr = f"{addr}/api/v2/write?org={org}&bucket={bucket}&precision={precision}"
             self.headers = {"Authorization": f"Token {token}"}
         else:
--- a/cryptofeed/backends/influxdb.py
+++ b/cryptofeed/backends/influxdb.py
@@ -18,7 +18,7 @@ LOG = logging.getLogger('feedhandler')
 
 
 class InfluxCallback(HTTPCallback):
-    def __init__(self, addr: str, db: str, key=None, create_db=True, numeric_type=str, **kwargs):
+    def __init__(self, addr: str, db=None, key=None, create_db=True, numeric_type=str, org=None, bucket=None, token=None, precision='ns', **kwargs):
         """
         Parent class for InfluxDB callbacks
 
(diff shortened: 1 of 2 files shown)


bmoscon:
@heartcored98 thanks for the PR! I'll take a look at it in the coming days. I'll need to test against influx 1.x which is use locally as well as install and test against 2.0

cryptofeed/backends/influxdb.py
```diff
@@ -53,17 +53,30 @@ def __init__(self, addr: str, db: str, key=None, create_db=True, numeric_type=st
...
+        bucket: str (For InfluxDB 2.0 compatibility)
+          Bucket name for authentication
+        token: str (For InfluxDB 2.0 compatibility)
+          Token string for authentication
+        precision: str (For InfluxDB 2.0 compatibility)
+          Precision level among (s, ms, us, ns) 
         """
         super().__init__(addr, **kwargs)
-        self.addr = f"{addr}/write?db={db}"
+        if (org and bucket and token):
```
bmoscon: you don't need ( ) around boolean logic in if statements in python

cryptofeed/backends/influxdb.py
```diff
@@ -53,17 +53,30 @@ def __init__(self, addr: str, db: str, key=None, create_db=True, numeric_type=st
...
+            self.addr = f"{addr}/api/v2/write?org={org}&bucket={bucket}&precision={precision}"
+            self.headers = {"Authorization": f"Token {token}"}
+        else:
+            if create_db:
```
bmoscon: in 2.0 do you no longer need to create the db?

cryptofeed/backends/influxdb.py
```diff
@@ -77,7 +90,7 @@ def __init__(self, addr: str, db: str, key=None, create_db=True, numeric_type=st
         d = d[:-1]
 
         update = f'{self.key}-{feed},pair={pair} {d},timestamp={timestamp}'
-        await self.http_write('POST', update)
+        await self.http_write('POST', update, self.headers)
```
bmoscon: you don't need to have a class member as a method argument, just access it in the method.

cryptofeed/backends/influxdb.py
heartcored98: No, db in the 1.x is replaced with a bucket in 2.0.

cryptofeed/backends/influxdb.py
heartcored98: In that case, will it better to create `header` inside the each method? 

Both `InfluxBookCallback._write_rows()` and `InfluxCallback.write()` need to pass the header dictionary for authentication when using influx 2.0.

One possible way is to always create the headers object inside the `_write_rows()` and `write()` before `await self.http_write()` method depends on the influx version instead of creating `self.headers`.  

Then, I think we need an initial argument with `InfluxCallback()` indicating whether the user would connect to influx 1.x or 2.0. 

If you give any guidelines or advice then I will modify it.

cryptofeed/backends/influxdb.py
bmoscon: do you need to create buckets before writing to them?

cryptofeed/backends/influxdb.py
bmoscon: ok, I see what you mean, you'd have to modify http_write in a way thats not ideal, so you should just leave that as-is.

cryptofeed/backends/influxdb.py
heartcored98: Following this [Influx 2.0 doc](https://v2.docs.influxdata.com/v2.0/organizations/buckets/create-bucket/), we can only create a bucket through a web GUI(it does not mean that http method is available) and CLI command.  

We should have to create bucket before writing, but it cannot be executed by http request.

bmoscon:
@heartcored98 I installed influx 2.0 beta and tried to test this but I cannot get it to work.

bmoscon:
@heartcored98 actually scratch that - its working now! Just need to test the changes against 1.7 now and then will merge

heartcored98:
Thank you for reviewing my pull request :)

PR #216, opened 2020-03-07
Message Receipt Timestamps

bmoscon:
This PR adds a second timestamp to all updates - the receipt timestamp. This indicates when the cryptostore feedhandler receives the message from the websocket. This timestamp is sometimes used to timestamp updates when no timestamp is present from the exchange. This will be the case when timestamp == receipt_timestamp

### DIFF
--- a/cryptofeed/backends/arctic.py
+++ b/cryptofeed/backends/arctic.py
@@ -33,9 +33,10 @@ class ArcticCallback:
         self.key = key if key else self.default_key
         self.numeric_type = numeric_type
 
-    async def write(self, feed, pair, timestamp, data):
+    async def write(self, feed, pair, timestamp, receipt_timestamp, data):
         df = pd.DataFrame({key: [value] for key, value in data.items()})
         df['date'] = pd.to_datetime(df.timestamp, unit='s')
+        df['receipt_timestamp'] = pd.to_datetime(df.receipt_timestamp, unit='s')
         df.set_index(['date'], inplace=True)
         df.drop(columns=['timestamp'], inplace=True)
         self.lib.append(self.key, df, upsert=True)
--- a/cryptofeed/backends/backend.py
+++ b/cryptofeed/backends/backend.py
@@ -12,49 +12,48 @@ from cryptofeed.backends._util import book_convert, book_delta_convert
 
 
 class BackendBookCallback:
-    async def __call__(self, *, feed, pair, book, timestamp):
-        data = {'timestamp': timestamp, 'delta': False, BID: {}, ASK: {}}
+    async def __call__(self, *, feed: str, pair: str, book: dict, timestamp: float, receipt_timestamp: float):
+        data = {'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': False, BID: {}, ASK: {}}
         book_convert(book, data, convert=self.numeric_type)
-        await self.write(feed, pair, timestamp, data)
+        await self.write(feed, pair, timestamp, receipt_timestamp, data)
 
 
 class BackendBookDeltaCallback:
-    async def __call__(self, *, feed, pair, delta, timestamp):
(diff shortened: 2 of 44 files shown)


cryptofeed/backends/arctic.py
```diff
@@ -33,9 +33,10 @@ def __init__(self, library, host='127.0.0.1', key=None, numeric_type=float, **kw
         self.key = key if key else self.default_key
         self.numeric_type = numeric_type
 
-    async def write(self, feed, pair, timestamp, data):
+    async def write(self, feed, pair, timestamp, receipt_timestamp, data):
         df = pd.DataFrame({key: [value] for key, value in data.items()})
         df['date'] = pd.to_datetime(df.timestamp, unit='s')
+        df['receipt_timestamp'] = pd.to_datetime(df.receipt_timestamp, unit='s')
```
PandaXcentric: is seconds granular enough here, or should it be 'ms'

cryptofeed/backends/backend.py
```diff
@@ -12,49 +12,48 @@
 
 
 class BackendBookCallback:
```
PandaXcentric: It looks like data always has `timestamp` and `receipt_timestamp`. I don't think your `write` functions need it explicitly passed since if it needs it, it's always accessible in the data

cryptofeed/backends/arctic.py
PandaXcentric: if you have receipt_timestamp explicitly passed, it probably makes more sense to use that rather than getting it from the data object.

I have another comment further down about passing receipt_timestamp explicitly though.

cryptofeed/exchange/binance.py
```diff
@@ -177,7 +179,7 @@ def _check_update_id(self, pair: str, msg: dict) -> (bool, bool):
             return
 
         delta = {BID: [], ASK: []}
-        timestamp = msg['E']
+        ts = msg['E']
```
PandaXcentric: I'd leave this as `timestamp` and rename the passed argument to `receipt_timestamp`. That way the naming of the variable is consistent all the way through (since you're passing `timestamp` into `receipt_timestamp` to `book_callback`

cryptofeed/exchange/binance.py
```diff
@@ -70,9 +70,10 @@ def _reset(self):
                                      side=SELL if msg['m'] else BUY,
                                      amount=amount,
                                      price=price,
-                                     timestamp=timestamp_normalize(self.id, msg['E']))
+                                     timestamp=timestamp_normalize(self.id, msg['E']),
+                                     receipt_timestamp=timestamp)
 
-    async def _ticker(self, msg):
+    async def _ticker(self, msg: dict, timestamp: float):
```
PandaXcentric: I'd call the variable receipt_timestamp since that's where you're passing it for consistency

cryptofeed/exchange/bitcoincom.py
```diff
@@ -46,24 +46,26 @@ def __reset(self):
                     }
                 ))
 
-    async def _trade(self, msg):
+    async def _trade(self, msg: dict, timestamp: float):
```
PandaXcentric: same comment here as the file above on the name of timestamp

cryptofeed/backends/backend.py
bmoscon: they all take the same args as the user defined callbacks, so it has to be in the args. Its a lot easier to just dump it in the data than to re-process the data with that info (in my opinion). some of them do just use it from the write as opposed to the data.

cryptofeed/backends/arctic.py
bmoscon: you're telling pandas what the data its formatting is in, the data is unix epoch time, so its in seconds, hence the 's'.

cryptofeed/exchange/binance.py
bmoscon: it also gets used as the timestamp if one is not available, so its still a timestamp

Issue #214, opened 2020-03-04
Feed Request - Upbit (Korean Exchange)


heartcored98:
Hi, I am really enjoying the cryptofeed library. 

However, I hope cryptofeed could support Upbit Exchange which is one of the largest exchange in the South Korea.

Here is the API doc [link](https://docs.upbit.com/docs/upbit-quotation-websocket). 

However, I am afraid documentation is only written in Korean. So I might implement Upbid feed myself as soon as possible.

bmoscon:
I'll definitely take a PR. Google translate seemed to translate the document just fine, so I could do it eventually. There is some documentation about adding a feed:

https://github.com/bmoscon/cryptofeed/blob/master/docs/exchange.md

its a little bit out of date, but not too much

bmoscon:
this was merged, so closing

PR #220, opened 2020-03-15
Upbit Feed Feature

heartcored98:
Hi, I implemented Upbit feed for #214  

Note that Upbit has following situation.

- Trade : fully support
- Orderbook : only support full fixed 15-depth orderbook (does not support delta)  
- Ticker : not support (ticker api does not cotain best_ask/best_bid price)

Implementing `_book` method was quite tricky since this is my first implementation of the feed.
I also added the `demo_upbit.py` for basic usage and add upbit case within `demo_book_delta.py`  

Hope this implementation is right.


cryptofeed/exchange/upbit.py
```diff
@@ -0,0 +1,210 @@
+import logging
+import json
+from decimal import Decimal
+import requests
+import zlib
```
bmoscon: you appear to not be using zlib or base64

examples/demo_book_delta.py
```diff
@@ -72,7 +72,7 @@ def check_books(self, master):
 
 def main():
     f = FeedHandler()
-
+
```
bmoscon: please trim/remove extra whitespace

cryptofeed/exchange/upbit.py
heartcored98: Yor are right, I will remove the zlib and base64

examples/demo_book_delta.py
heartcored98: Okay, no problem

heartcored98:
For notice purpose, I also added get_active_symbol method. 

I hope you to check these changes ~

bmoscon:
@heartcored98 thanks for the PR! I'll review and test it later today.

joelrich:
@heartcored98 Please have look here: https://github.com/bmoscon/cryptofeed/issues/232#issuecomment-614546881
I am pretty sure the order book implementation for Upbit is not correct. It might be that it does not remove stale order book entries when the price is moving.

heartcored98:
@joelrich Okay, I will look at the issue as soon as possible. Thank you for notice.

PR #227, opened 2020-04-02
Support DSX exchange

vincentmele:
I have added support for DSX exchange, https://dsxglobal.com. I added a test for symbols but it seems to be failing and I cannot determine why. The troublesome pair (BTCUSDT) works fine when streaming data, as seen in examples/demo_dsx.py.

Please let me know if there is anything else you need for this to be accepted.


bmoscon:
Awesome! Thanks for the contribution - I'll take a look and merge later tonight. I can help fix/debug the issues too.

cryptofeed/exchange/dsx.py
```diff
@@ -0,0 +1,114 @@
...
+associated with this software.
+'''
+import json
+import logging
+from decimal import Decimal
+
+from sortedcontainers import SortedDict as sd
+
+from cryptofeed.feed import Feed
+from cryptofeed.defines import TICKER, L2_BOOK, TRADES, BUY, SELL, BID, ASK, DSX
```
bmoscon, line 14: probably should import DSX here as DSX_id, or _DSX, etc.

```python
from cryptofeed.defines import TICKER, L2_BOOK, TRADES, BUY, SELL, BID, ASK, DSX as DSX_id
```

then

```python
class DSX(Feed):
    id = DSX_id
```

vincentmele: corrections made and pushed

bmoscon:
@vincentmele sorry to let you know, but it seems theyve updated their API and websocket is no longer supported. I've had to remove this exchange :(

PR #233, opened 2020-04-07
Fix demo_okex_futures.py

xiandong79:


bmoscon:
thanks for the PR @xiandong79 !

Issue #231, opened 2020-04-06
Change json library to yapic for performance


vincentmele:
I have to say, thank you for making this easy. I dumped raw json data using examples/demo_raw_data.py, and then used it in the test below. The results show yapic is about 1.3x to 2.0x the speed of the standard library for parsing l2_book, ticker, and trades data from 100 MB files from Binance and Deribit feeds.

Based on these results, I am going to proceed with integrating yapic for further stability testing.

```
binance results
file lines: 211316
yapic: 13.38 seconds
json: 23.62 seconds
yapic lps: 47390.76 lines per second
json lps: 26838.02 lines per seond
json/yapic: 1.77x

deribit results
file lines: 310000
yapic: 18.85 seconds
json: 37.63 seconds
yapic lps: 49332.11 lines per second
json lps: 24717.06 lines per seond
json/yapic: 2.00x

```

```
import joblib
import timeit

iterations = 3
files =  { 'binance' : 'BINANCE12f2ac65-b5c5-404d-b8b4-8f22c04da3a4.0',
           'deribit' : 'DERIBIT252a6e79-7a32-4038-a64a-b1f7b29d6837.3' }

def load_file(filepath):
  output = []
  with open(filepath) as fp:
    line = fp.readline()
    while line:
      output.append(line.split(' ')[1])
      line = fp.readline()
  joblib.dump(output, 'json.pkl')
  return len(output)

for k, f in files.items():
  length = load_file(f)
  yapic_time = timeit.timeit('[json.loads(msg,parse_float=Decimal) for msg in txt]',setup="from yapic import json; import joblib; from decimal import Decimal; txt = joblib.load('json.pkl')", number=iterations)
  yapic_lps = length * iterations / yapic_time

  json_time = timeit.timeit('[json.loads(msg,parse_float=Decimal) for msg in txt]',setup="import json; import joblib; from decimal import Decimal; txt = joblib.load('json.pkl')", number=iterations)
  json_lps = length * iterations / json_time

  ratio = json_time / yapic_time

  print(f'{k} results')
  print(f'file lines: {length}')
  print(f'yapic: {yapic_time:.2f} seconds')
  print(f'json: {json_time:.2f} seconds')
  print(f'yapic lps: {yapic_lps:.2f} lines per second')
  print(f'json lps: {json_lps:.2f} lines per seond')
  print(f'json/yapic: {ratio:.2f}x')

```

[...]

PR #237, opened 2020-04-14
Use yapic.json instead of json for 1.5x to 2x performance gain

vincentmele:
Documented in #231 .

Tested for a week continuously (on only a couple exchanges, including Huobi) and I see no abnormal results.

### DIFF
--- a/cryptofeed/exchange/huobi.py
+++ b/cryptofeed/exchange/huobi.py
@@ -74,7 +74,7 @@ class Huobi(Feed):
 
     async def message_handler(self, msg: str, timestamp: float):
         # unzip message
-        msg = zlib.decompress(msg, 16 + zlib.MAX_WBITS).decode(encoding='utf-8', errors='strict')
+        msg = zlib.decompress(msg, 16 + zlib.MAX_WBITS)
         msg = json.loads(msg, parse_float=Decimal)
 
         # Huobi sends a ping evert 5 seconds and will disconnect us if we do not respond to it
--- a/cryptofeed/exchange/huobi.py
+++ b/cryptofeed/exchange/huobi.py
@@ -5,7 +5,7 @@ Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
 import logging
-import json
+from yapic import json
 from decimal import Decimal
 import zlib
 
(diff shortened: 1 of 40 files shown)


bmoscon:
you have the wrong dependency here. It should be:

https://pypi.org/project/yapic.json/

So, i dont think you need to re-test for a week, but thats definitely the wrong library. yapic installs tensoflow and a million other reqs that arent needed

vincentmele:
> you have the wrong dependency here. It should be:
> 
> https://pypi.org/project/yapic.json/
> 
> So, i dont think you need to re-test for a week, but thats definitely the wrong library. yapic installs tensoflow and a million other reqs that arent needed

Thanks for catching the error in setup.py. We don't need tensorflow (yet) :-)

cryptofeed/exchange/huobi.py
```diff
@@ -74,7 +74,7 @@ async def _trade(self, msg: dict, timestamp: float):
 
     async def message_handler(self, msg: str, timestamp: float):
         # unzip message
-        msg = zlib.decompress(msg, 16 + zlib.MAX_WBITS)
+        msg = zlib.decompress(msg, 16 + zlib.MAX_WBITS).decode(encoding='utf-8', errors='strict')
```
bmoscon, line 77: looks like the library has been updated to do byte parsing, so you should be able to remove this and require the latest version of the library in setup.py

dikkechill:
Good idea to improve the json decoding. How does yapic.json perform compared to e.g. ultrajson?
See https://github.com/ultrajson/ultrajson for some benchmarks with other implementations.

vincentmele:
> Good idea to improve the json decoding. How does yapic.json perform compared to e.g. ultrajson?
> See https://github.com/ultrajson/ultrajson for some benchmarks with other implementations.

ultrajson doesn't support "parse_float=decimal.Decimal", which is required. I opened the issue here: https://github.com/ultrajson/ultrajson/issues/401, then found that yapic supported it and was extremely fast.

bmoscon:
@dikkechill yep - the big requirement here is parsing floats to decimal.Decimal, I haven't found any other libs that support this, but I also haven't exhaustively searched either

Issue #232, opened 2020-04-07
l2 order book data for some exchanges is inconsistent (ask and bid prices overlap)


joelrich:
I have written a function to reconstruct order books from recorded order book data and found that it is inconsistent for some exchanges.

This is how I persist the data:
```python
            batch_row = dict(timestamp=timestamp, delta=delta)
            for side in [BID, ASK]:
                entries = data.get(side, [])
                if isinstance(entries, dict):
                    entries = entries.items()

                batch_row[f"{side}_prices"] = " ".join(to_str(price) for price, size in entries)
                batch_row[f"{side}_sizes"] = " ".join(to_str(size) for price, size in entries)
            self.writer.append(self.exchange, pair, L2_BOOK + "_batch", store_timestamp, batch_row)
```
When I read the data, I perform the following check:
```python
        nonzero_bid_prices = col_bid_prices[i][col_bid_sizes[i] > 0]
        nonzero_ask_prices = col_ask_prices[i][col_ask_sizes[i] > 0]
        if min(len(nonzero_bid_prices), len(nonzero_ask_prices)) > 0:
            assert np.max(nonzero_bid_prices) < np.min(nonzero_ask_prices), "data inconsistent"
```
This check ensures that the snapshot/delta does not contain overlapping prices.
Unfortunately, it fails for some exchanges, e.g. Bitmex. Here is an example:

```
Time: 2020-03-12 23:42:21.158979416
Delta: False
Bid prices: 0.5 1 1.5 2 3 3.5 4 5 5.5 1 10.5 11.5 16.5 23 31 32 33 ... 4630.5 4631 4631.5 4633 4633.5 4634.5
Ask prices: 464 4640.5 4641.5 4642 4642.5 4644.5 4645 4645.5 ... 600884.5 666666 993 999998 999999 1
```
There are two problematic ask prices: 464 (this is not a typo) and 1 (after 999999)

To prevent such problems, I think a similar check should run after retrieving order book data. If asks and bids overlap, the order book should be discarded followed by a reconnect/re-initialization of the order book.

joelrich:
@bmoscon what is your opinion on this issue? Personally, I find it worrisome that there is no guarantee for some weak consistencies (e.g. bid and ask do not overlap). If only a single websocket message is lost for whatever reason, this may affect all snapshots until the process is restarted.

bmoscon:
[...]

PR #240, opened 2020-04-25
Throw exception if bid and ask are overlapping

joelrich:
See https://github.com/bmoscon/cryptofeed/issues/232 for discussion
This will completely break the Upbit feed if L2_BOOK is enabled (most likely since either the implementation is wrong or the exchange notifications are insufficient).

### DIFF
--- a/cryptofeed/feed.py
+++ b/cryptofeed/feed.py
@@ -6,11 +6,14 @@ associated with this software.
 '''
 import uuid
 from collections import defaultdict
+from sortedcontainers import SortedDict as sd
+
 
 from cryptofeed.callback import Callback
 from cryptofeed.standards import pair_std_to_exchange, feed_to_exchange, load_exchange_pair_mapping
 from cryptofeed.defines import TRADES, TICKER, L2_BOOK, L3_BOOK, VOLUME, FUNDING, BOOK_DELTA, OPEN_INTEREST, BID, ASK
 from cryptofeed.util.book import book_delta, depth
+from cryptofeed.exceptions import BidAskOverlapping
 
 
 class Feed:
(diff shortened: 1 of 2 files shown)


cryptofeed/feed.py
```diff
@@ -101,12 +105,22 @@ async def book_callback(self, book: dict, book_type: str, pair: str, forced: boo
...
+        self.check_bid_ask_overlapping(book, pair)
         if book_type == L2_BOOK:
             await self.callback(L2_BOOK, feed=self.id, pair=pair, book=book, timestamp=timestamp, receipt_timestamp=receipt_timestamp)
         else:
             await self.callback(L3_BOOK, feed=self.id, pair=pair, book=book, timestamp=timestamp, receipt_timestamp=receipt_timestamp)
         self.updates[pair] = 0
 
+    def check_bid_ask_overlapping(self, book, pair):
+        bid, ask = book[BID], book[ASK]
+        assert isinstance(bid, sd) and isinstance(ask, sd)
```
bmoscon, line 118: I'm thinking we don't need this as every exchange/feed implementation uses sorted dict.

joelrich: Removed

bmoscon:
@joelrich thanks - the build is broken so I'm going to fix that before taking in any more changes. Should be merged shortly

bmoscon:
thanks for the PR. Merged. I added a flag so its optionally enabled when the feed is configured so I can test out different settings on different feeds to try and narrow down a root cause. Could just be lossy exchanges, want to do more digging.

joelrich:
In my log files it has happened for the following exchanges so far:

226 BITTREX   
98 BITFINEX  
77 BITSTAMP  
12 KRAKEN    
7 OKEX     
2 BITMEX

PR #241, opened 2020-04-25
Upbit book implementation fix

vdbelt:
As reported by @joelrich in https://github.com/bmoscon/cryptofeed/pull/240


joelrich:
Does this remove zero entries from the order book?

vdbelt:
Yes, it filters out zero entries from the book.

joelrich:
@vdbelt Ok, thanks, now it makes more sense. I assume you have checked that bid and ask are no more overlapping?

vdbelt:
Yes, I tested it on a branch with your outstanding PR applied. No more overlapping. 


I found another issue however when book deltas are enabled and max_depth is specified. Deltas would be provided disregarding the max_depth. Happens to the Huobi feeds as well. Seems to be an edge case that needs to be handled by the callback in feed.py. Will look at that.

PR #242, opened 2020-05-03
Integrate Blockchain Exchange

simon-bc:
This PR integrates Blockchain's exchange, https://exchange.blockchain.com/

Not sure what to do with the ticker update - the websocket ticker updates the last traded price, rather than bid-ask prices. Any feedback would be much appreciated.

Let me know if there's anything else needed 👍


cryptofeed/feed.py
```diff
@@ -52,6 +54,7 @@ def __init__(self, address, pairs=None, channels=None, config=None, callbacks=No
                           VOLUME: Callback(None),
                           FUNDING: Callback(None),
                           OPEN_INTEREST: Callback(None),
+                          BOOK_DELTA: Callback(None),
```
vdbelt, line 57: Delta callbacks aren't called by the feed implementation itself. `BOOK_DELTA` should't be called unless a callback is actually passed to the feed. So a default should not be necessary.

cryptofeed/standards.py
```diff
@@ -174,6 +177,7 @@ def timestamp_normalize(exchange, ts):
         BINANCE_US: 'ticker',
         BINANCE_JERSEY: 'ticker',
         BINANCE_FUTURES: UNSUPPORTED,
+        BLOCKCHAIN: 'ticker',
```
vdbelt, line 180: Since bid/ask prices aren't provided by the feed, just leave this as `UNSUPPORTED` to avoid confusion.

cryptofeed/feed.py
```diff
@@ -119,7 +123,6 @@ async def apply_depth(self, book: dict, do_delta: bool, pair: str):
             self.previous_book[pair] = ret
             return delta, ret
 
-        delta = []
```
vdbelt, line 122: Is this supposed to be here?

cryptofeed/feed.py
simon-bc: is it not called from inside `cryptofeed.feed.Feed#book_callback`  ?

or is that part only every reached if `self.do_deltas = True`

cryptofeed/standards.py
simon-bc: yeah - good point 👍

cryptofeed/feed.py
simon-bc: it is defined as the output of the function below, so it serves no real purpose

cryptofeed/feed.py
bmoscon: you don't need to put book deltas in here, you can remove this line

PR #247, opened 2020-06-12
Ftx funding rest api added

Quenos:
API and test files created

### DIFF
--- a/cryptofeed/rest/ftx.py
+++ b/cryptofeed/rest/ftx.py
@@ -64,7 +64,60 @@ class FTX(API):
         for data in self._get_trades_hist(symbol, start, end, retry, retry_wait):
             yield data
 
-    def _dedupe(self, data, last):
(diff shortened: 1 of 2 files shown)


bmoscon:
@Quenos thanks for the PR. This won't resolve the issue you commented on though, as the code here in Rest is mainly for collecting historic data. If you want to add funding data to FTX you need to create a coroutine in the websocket code that periodically (based on rest API limits) checks the rest endpoint for the latest funding data, and if its new/changed, updates the funding callback. There is an example of an exchange that works like this Coinbene, in the code (if you want to do it, if not, no worries)

Quenos:
@bmoscon Yes I found that out as well. For my purposes this is enough, but I started to also create it more in line with Coinebene.
I have it working now as a separate exchange, but would like to integrate it in the FTX exchange class such that the user can registrate L2_BOOK and FUNDING in the same add_feed call. But it needs some bug fixing, because as soon as FUNDING is added the websocket interface stops working.

PR #249, opened 2020-06-14
added stream for FTX funding 

Quenos:


bmoscon:
@Quenos I'll take this as is, but mostly rewrite a lot of it. Is thats ok with you, I'll merge this. If not then here is what needs to change - you dont need a new feed class. You can add support for this in the existing FTX class. If funding is enabled, you just have to set up a coroutine that triggers N times a minute (whatever is allowable by the rest api limits) to call that endpoint and trigger the callback

Quenos:
Ok, but just to make sure. The code I've added to feedhandler.py needs to stay to filter out the request to get a funding feed, right?
(logic will probably change of course)

Quenos:
@bmoscon and of course it's ok if you change it ;)

Issue #256, opened 2020-06-26
Kraken order book updates broken: not enough values to unpack (expected at least 2, got 1)


joelrich:
Error:
```
    price, size, *_ = update
ValueError: not enough values to unpack (expected at least 2, got 1)
2020-06-26 11:24:06,794 : ERROR : KRAKEN: encountered an exception, reconnecting
```

They seem to have changed the messages. E.g.:
```
m: {'b': [['1.024871', '655.16472904', '1593163445.981705']], 'c': '2712626796'}
```
`c` is then interpreted as `ASK` instead of being ignored.

pyrex41:
Fix suggested in #258

bmoscon:
fixed

PR #258, opened 2020-06-30
fix for Kraken update

pyrex41:
Fix for issue identified by Issue #256 . Seems to have only applied to book updates via websocket.

### DIFF
--- a/cryptofeed/exchange/kraken.py
+++ b/cryptofeed/exchange/kraken.py
@@ -97,21 +97,26 @@ class Kraken(Feed):
         else:
             for m in msg:
                 for s, updates in m.items():
-                    side = BID if s == 'b' else ASK
-                    for update in updates:
-                        price, size, *_ = update
-                        price = Decimal(price)
-                        size = Decimal(size)
-                        if size == 0:
-                            # Per Kraken's technical support
-                            # they deliver erroneous deletion messages
-                            # periodically which should be ignored
-                            if price in self.l2_book[pair][side]:
-                                del self.l2_book[pair][side][price]
-                                delta[side].append((price, 0))
-                        else:
-                            delta[side].append((price, size))
-                            self.l2_book[pair][side][price] = size
+                    side = False
+                    if s == 'b':
+                        side = BID
+                    elif s == 'a':
+                        side = ASK
+                    if side:
+                        for update in updates:
+                            price, size, *_ = update
+                            price = Decimal(price)
+                            size = Decimal(size)
+                            if size == 0:
+                                # Per Kraken's technical support
+                                # they deliver erroneous deletion messages
+                                # periodically which should be ignored
+                                if price in self.l2_book[pair][side]:
+                                    del self.l2_book[pair][side][price]
+                                    delta[side].append((price, 0))
+                            else:
+                                delta[side].append((price, size))
+                                self.l2_book[pair][side][price] = size
             for side in (BID, ASK):
                 while len(self.l2_book[pair][side]) > self.book_depth:
                     del_price = self.l2_book[pair][side].items()[0 if side == BID else -1][0]


cryptofeed/exchange/kraken.py
```diff
@@ -97,21 +97,26 @@ async def _book(self, msg: dict, pair: str, timestamp: float):
...
-                                delta[side].append((price, 0))
-                        else:
-                            delta[side].append((price, size))
-                            self.l2_book[pair][side][price] = size
+                    side = False
+                    if s == 'b':
+                        side = BID
+                    elif s == 'a':
+                        side = ASK
+                    if side:
```
bmoscon, line 105: will side ever be false?

pyrex41: Yes, I made this change specifically for that case. Kraken's updates now include a checksum with each (see #256, and https://docs.kraken.com/websockets/#book-checksum) where s == 'c'. This change was made to ignore the dictionary entries for the checksum, and to just process the bid/ask data.

pyrex41: Quick test script for convenience

`from decimal import Decimal

from cryptofeed.callback import TickerCallback, TradeCallback, BookCallback, FundingCallback
from cryptofeed.exchanges import Kraken
from cryptofeed.defines import L2_BOOK

from pprint import pprint


# Examples of some handlers for different updates. These currently don't do much.
# Handlers should conform to the patterns/signatures in callback.py
# Handlers can be normal methods/functions or async. The feedhandler is paused
# while the callbacks are being handled (unless they in turn await other functions or I/O)
# so they should be as lightweight as possible

async def book(feed, pair, book, timestamp, receipt_timestamp):
    if type(book) == dict:
        pprint(book.keys())
    else:
        pprint(book)
    print(f'Timestamp: {timestamp} Feed: {feed} Pair: {pair} Receipt_Timestamp: {receipt_timestamp}')

def main():
    f = FeedHandler()
    # Note: EXX is extremely unreliable - sometimes a connection can take many many retries
    # f.add_feed(EXX(pairs=['BTC-USDT'], channels=[L2_BOOK, TRADES], callbacks={L2_BOOK: BookCallback(book), TRADES: TradeCallback(trade)}))
    f.add_feed(Kraken(config={L2_BOOK: ['BTC-USD']}, callbacks={L2_BOOK: BookCallback(book)}))
    f.run()


if __name__ == '__main__':
    print('starting')
    main()
`

pyrex41: Quick test script for convenience, 

```python
from decimal import Decimal

from cryptofeed.callback import TickerCallback, TradeCallback, BookCallback, FundingCallback
from cryptofeed.exchanges import Kraken
from cryptofeed.defines import L2_BOOK

from pprint import pprint


# Examples of some handlers for different updates. These currently don't do much.
# Handlers should conform to the patterns/signatures in callback.py
# Handlers can be normal methods/functions or async. The feedhandler is paused
# while the callbacks are being handled (unless they in turn await other functions or I/O)
# so they should be as lightweight as possible

async def book(feed, pair, book, timestamp, receipt_timestamp):
    if type(book) == dict:
        pprint(book.keys())
    else:
        pprint(book)
    print(f'Timestamp: {timestamp} Feed: {feed} Pair: {pair} Receipt_Timestamp: {receipt_timestamp}')

def main():
    f = FeedHandler()
    # Note: EXX is extremely unreliable - sometimes a connection can take many many retries
    # f.add_feed(EXX(pairs=['BTC-USDT'], channels=[L2_BOOK, TRADES], callbacks={L2_BOOK: BookCallback(book), TRADES: TradeCallback(trade)}))
    f.add_feed(Kraken(config={L2_BOOK: ['BTC-USD']}, callbacks={L2_BOOK: BookCallback(book)}))
    f.run()


if __name__ == '__main__':
    print('starting')
    main()
```

PR #268, opened 2020-08-02
Rephrase INSTALL.md + Add isort settings + Some minor changes

olibre:
* Rephrase INSTALL.md
    * Use better English-style sentences
    * Add more explanations in INSTALL.md
* Improve README.md
    * Use correct exchange / market place names
    * Sort exchange name alphabetically
* Reformat setup.py (source code) to match black (Python formatter)
* Add isort settings having same result as black
* Disable Pylint in Pipfile because it downgrades isort to version 4 not producing same results as black
    * However isort settings are backward compatible with v4
    * Thus Pylint can be enabled/used by anyone
    * Will wait for Pylint-2.6 to enable it again in Pipfile


README.md
```diff
@@ -51,33 +51,34 @@ To see an example of an application using cryptofeed to aggregate and store cryp
...
+* Binance Jersey
+* Binance US
+* BitMEX
+* CoinBene
+* Coinbase (GDAX)
 * Deribit
 * EXX
 * FTX
-* FTX US
+* FtxUs
```
bmoscon, line 71: If I go to the exchange's website, its all caps

bmoscon:
also, I should note I HATE black - I really disagree with a lot of the formatting changes it makes, so if you enable it and code starts changing, that will be an issue. The changes in setup.py don't really matter to me, but the bulk of the library code does.

olibre:
### FTX US

You are right for FTX US https://ftx.us/
I apologize: I really don't know why I thought it was FtxUs !
Thanks for checking ;-)

### `isort` / `black` formatter

I appreciate it when a tool helps to keep the source code clean.
But just like you, I disagree with many of Black rules, and I don't use it for my projects.

However I appreciate `isort` and some other tools as `autopep8`, `atoflake`, `unify` and `docformatter` (I do not use `yapf`).

For the `isort` configuration, I thing the *uncompromising* `black` style for `import` section is fine.
Instead of debating about the `isort` settings, we can just use its `black` profile (starting from `isort` v5).

### setup.py

I accidentally reformatted `setup.py` while checking that my `isort` settings performs as well as `black` (checking `isort` v4 and v5). I liked the `setup.py` reformatted by `black`, and I added it to the Pull Request.

### Fixing

I am fixing README.md:  FtxUs -> FTX US

### Feedback

Would you like me to change something else?

olibre:
Hi @bmoscon 
I just reverted FtxUs -> FTX US.
I have also checked the names of the other exchanges / market places.
During the check on the official websites => I also added these links.
I hope you think this is a good idea.

README.md
```diff
@@ -51,33 +51,34 @@ To see an example of an application using cryptofeed to aggregate and store cryp
...
-* Gemini
-* HitBTC
-* Huobi
-* HuobiDM
-* Kraken
-* Kraken Futures
-* OKCoin
-* OKEx
-* Poloniex
-* Upbit
```
bmoscon, line 80: why was upbit removed?

olibre: Good point.
To answer your question I suppose I dropped Upbit during my copy/past.
I remember having counted the number of exchanges before and after and found the same number...
OK I understand my mistake: I added Huobi Swap 🤦 
Thanks again for your review :-D

Note: I have also added the exchange names within the Pypi project keywords in `setup.py`:
https://github.com/olibre/cryptofeed/blob/clean-import-sections/setup.py#L45
The Upbit exchange was already there.

PR #269, opened 2020-08-02
Format "import" section in folder "examples" only

olibre:
Hi @bmoscon 

What do you think about this cleaning PR?
Do you like the `isort` result?
It sorts the imports alphabetically and when the line is too long, splits it using parenthesis.

On my opinion, `isort` does a good job.

If you like it, I can apply it on the entire code base.

### DIFF
--- a/examples/demo.py
+++ b/examples/demo.py
@@ -6,11 +6,45 @@ associated with this software.
 '''
 from decimal import Decimal
 
-from cryptofeed.callback import TickerCallback, TradeCallback, BookCallback, FundingCallback
 from cryptofeed import FeedHandler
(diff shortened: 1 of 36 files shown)


bmoscon:
sorting them is fine, but the line length seems off. How many chars does it consider "too long" ?

olibre:
The line length is define by the parameter `line_length` 
https://github.com/bmoscon/cryptofeed/blob/master/pyproject.toml#L19
The value `line_length=88` is the default for `black`.
Please see my last commit with the value `line_length=122`

I am feeling you dislike when there are so much lines per import.
You may want a more compact `isort` result...
I am preparing another Pull Request that you may prefer...

olibre:
Hi @bmoscon 

I removed the `black` profile in the `isort` settings and set `line_length=120` to obtain a compact result as I think you prefer.

In this Pull Request we keep the file `pyproject.toml` because this a good way to store the `line_length=120` and we can also share some tips in the file comments. This file `pyproject.toml` will probably be re-used in the coming years to contain the settings of other Python tools.

I hope this Pull Request meets your requirements.

PR #273, opened 2020-08-04
Fix typos in CHANGES.md

olibre:
* ~~The file `backends/mongo.py` imports bson => Add this missing dependency~~
* Fix some typos in CHANGES.md
* Use option --upgrade consistency in Travis file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setup(
         "redis": ["aioredis"],
         "arctic": ["arctic"],
         "zmq": ["pyzmq"],
-        "mongo": ["motor"],
+        "mongo": ["bson", "motor"],
         "kafka": ["aiokafka"],
         "rabbit": ["aio_pika", "pika"],
         "postgres": ["asyncpg"],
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -74,7 +74,7 @@ def run_tests(self):
         "redis": ["aioredis"],
         "arctic": ["arctic"],
         "zmq": ["pyzmq"],
-        "mongo": ["motor"],
+        "mongo": ["bson", "motor"],
```
bmoscon, line 77: this doesnt need to be manually installed. motor includes it. Try creating a virtualenv, and installing only motor - you'll find that bson is available for import. Its built into pymongo/motor

olibre: You are right, I can read from PyMongo packages distribution:
https://github.com/mongodb/mongo-python-driver#installation

> Do not install the "bson" package from pypi. PyMongo comes with its own bson package; doing "easy_install bson" installs a third-party package that is incompatible with PyMongo.

olibre:
I was not aware that a Python distribution could installs multiple packages at the root level.
Thank you for for eyes 👀👍 

I have reverted the main feature of this PR, the rest is minor, but you may merge it:

* Fix some typos in CHANGES.md
* Use option --upgrade consistency in Travis file
* Remove an unnecessary comment in Pipfile

PR #274, opened 2020-08-05
Add active option contracts retrieval to OKEx

rhotchkiss:
Have added options to OKEx in a similar manner to the futures. Happy to discuss


examples/demo_okex_futures.py
```diff
@@ -1,6 +1,5 @@
 '''
 Copyright (C) 2017-2020  Bryant Moscon - bmoscon@gmail.com
-
```
bmoscon, line 3: can you please replace that line

rhotchkiss: My bad; done & done

PR #275, opened 2020-08-05
Use `async` and `await` in API Rest

olibre:
Hi @bmoscon 

This PR to boost the performance of the API Rest by using good AsyncIO practices :
- Use `async` and `await` in API Rest
- Replace `time.sleep()` --> `await asyncio.sleep()`
- Adapt the unit tests with `@pytest.mark.asyncio`

Moreover, all the functions `ticker()`, `trades()` and `funding()` use the parameters `start=None` and `end=None` because on some exchanges, these parameters were already present, but not within the base `class Api`. Now everything is consistent.

Therefore the parameters `start_date` and `end_date` in `Ftx.funding()` have been renamed.
And the nested `helper()` function does not need these parameters because it already accesses scoped variables of the parent function.

Please perform a carefully code review.
Cheers

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,7 +16,7 @@ pandas = "*"
 requests = ">=2.18.4"
 sortedcontainers = ">=1.5.9"
 websockets = ">=7.0"
-"yapic.json" = ">=1.4.3"   # fast, drop-in replacement of the built-in python.json
+"yapic.json" = ">=1.4.3"   # fast drop-in replacement of the built-in python.json
 # --- Optional packages to speed-up Cryptofeed -----------------------------------------------
 aiodns = ">=1.1"   # aiodns speeds up DNS resolving
 cchardet = "*"     # cchardet is a faster replacement for chardet
--- a/cryptofeed/rest/api.py
+++ b/cryptofeed/rest/api.py
@@ -88,19 +88,19 @@ class API:
             resp.raise_for_status()
 
     # public / non account specific
-    def ticker(self, symbol: str, retry=None, retry_wait=10):
+    async def ticker(self, symbol: str, start=None, end=None, retry=None, retry_wait=10):
         raise NotImplementedError
 
-    def trades(self, symbol: str, start=None, end=None, retry=None, retry_wait=0):
+    async def trades(self, symbol: str, start=None, end=None, retry=None, retry_wait=0):
         raise NotImplementedError
 
-    def funding(self, symbol: str, retry=None, retry_wait=0):
+    async def funding(self, symbol: str, start=None, end=None, retry=None, retry_wait=10):
         raise NotImplementedError
 
-    def l2_book(self, symbol: str, retry=None, retry_wait=0):
+    async def l2_book(self, symbol: str, retry=None, retry_wait=0):
         raise NotImplementedError
 
-    def l3_book(self, symbol: str, retry=None, retry_wait=0):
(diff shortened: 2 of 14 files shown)


bmoscon:
Thats a pretty big change, and it will also break other applications I have that use this. I'm not saying I won't take it, but I'll need to coordinate this change with others

olibre:
Hi @bmoscon 
To avoid a break of the Cryptofeed API, we can expose the API function twice: 
1. the synchronous functions keep the same name
2. the asynchronous ones with prefix `async_`

There are some Python libraries having both synchronous and asynchronous API.

### Blocking functions calling the asyncio ones

An example is Playwright, see the documentation: 
https://github.com/microsoft/playwright-python#pytest

The blocking functions call the asyncio one via the `class SyncBase` that is implemented here:
https://github.com/microsoft/playwright-python/blob/master/playwright/sync_base.py#L88

You may deprecate the blocking functions drop them after some releases (after some years) : keep Cryptofeed clean. Or you can decide to support both API for longer time...

Nevertheless, converting a synchronous call to an asynchronous call is lots of work (and testing) and I do not want to do it alone. If you are interested by that way, please consider doing this with me.

### Copy-past the entire `rest` folder into `async_rest`

However there is another simpler way to provide both synchronous and asynchronous API: Keep the current `rest` folder unchanged, and make the changes (this PR) within a copy in `async_rest`.

I my opinion, the trend is to use more and more asyncio for such libraries, and most Cryptofeed-based applications will migrate in the coming year. 

So to avoid double maintenance between the two folders, I recommend to let the synchronous `rest` folder die, without evolution, just major bug fixes. And after some years, when the exchanges used in the `rest` folder will be broken (dead), you will remove it. If some users really want to benefit from the new features in `async_rest`, they can do the maintenance and send you PRs.

### Your choice ?

1. Synchronous API only, and you reject my current PR
2. Asynchronous API only, and you synchronize the next major Cryptofeed delivery with other projects
3. Synchronous API that calls the asynchronous API
4. Copy `rest` into the new folder `async_rest` and leave `rest` diying
5. Copy `rest` into the new folder `async_rest` and maintain both (at least for a few of months/years)

My personal favorite is number 4.
What about you?

bmoscon:
@olibre I think we should just have both supported. I think the idea of prefixing (or postfixing) the method names with async is sufficient.

olibre:
Hi @bmoscon 

I see two ways to support both Synchronous / Asynchronous APIs:
1. The Synchronous API calls the Asynchronous API using quite complex wrapper (see project [Playwright](https://github.com/microsoft/playwright-python/blob/master/playwright/sync_base.py#L88))
2. Duplicate the `rest` folder (new `async_rest` folder) and maintain both folders for months/years

Complexity:
1. The first one is difficult to develop in the beginning but will ease maintenance
2. The second one is very simple but more work to mirror one change from one folder to the other one

What is your option?

bmoscon:
I prefer option 3. Put them in the same classes

bmoscon:
@olibre since you never responded - should I assume you do not like my preference? I'm happy to do the work if you like, but I want to avoid having two separate classes or modules

olibre:
Hi @bmoscon 
I was on holidays.
Yes you can start working on the Sync/Async API yourself 👍  
I have not yet worked on this subject.
Please let me know when you are ready for a peer review :)

bmoscon:
given that this effort will be done in a separate branch/PR closing this

PR #283, opened 2020-08-23
Require aiofile>=2.0.0; fix write bug in AsyncFileCallback's write method for aiofile>=2.0.0 

### LINKED ISSUE #282: AsyncFileCallback fails when `length` is reached due to change in aiofile>=2.0.0
**Describe the bug**
The `AsyncFileCallback` in async_file.py relies on [the aiofile package](https://github.com/mosquito/aiofile). When aiofile>=2.0.0 is installed, and the number of writes to the callback reaches the `length` argument provided to an `AsyncFileCallback`, a TypeError is raised.

**To Reproduce**
Install aiofile>=2.0.0. Install cryptofeed. Run the following mwe:
```
from cryptofeed.callback import TradeCallback
from cryptofeed.defines import TRADES
from cryptofeed.exchanges import Coinbase
from cryptofeed.util.async_file import AsyncFileCallback

async def trade(feed, pair, order_id, timestamp, side, amount, price, receipt_timestamp):
    print(f"{timestamp}: Trade occurred")

def main():
    f = FeedHandler(raw_message_capture=AsyncFileCallback('./', length=10))
    f.add_feed(Coinbase(pairs=['BTC-USD'], channels=[TRADES], callbacks={TRADES: TradeCallback(trade)}))
    f.run()

if __name__ == '__main__':
    main()
``` 
This results in the error and traceback after 9 trades have occurred:
```
2020-08-23 14:18:45,985 : ERROR : COINBASE: encountered an exception, reconnecting
Traceback (most recent call last):
  File "/home/g/.local/lib/python3.8/site-packages/cryptofeed/feedhandler.py", line 211, in _connect
    await self._handler(websocket, feed.message_handler, feed.uuid)
  File "/home/g/.local/lib/python3.8/site-packages/cryptofeed/feedhandler.py", line 231, in _handler
    await self.raw_message_capture(message, self.last_msg[feed_id], feed_id)
  File "/home/g/.local/lib/python3.8/site-packages/cryptofeed/util/async_file.py", line 42, in __call__
    await self.write(uuid)
  File "/home/g/.local/lib/python3.8/site-packages/cryptofeed/util/async_file.py", line 32, in write
    self.pointer[uuid] += len(r)
TypeError: object of type 'int' has no len()
```
Changing the `length` argument provided to `AsyncFileCallback` will change the number of trades at which the exception occurs.

**Expected behavior**
No error

**Operating System:**
Ubuntu Server 20.04

**Cryptofeed Version**
1.5.0

**Cause**
This bug is due to a change in aiofile. In aiofile>=2.0.0, a new library [caio](https://github.com/mosquito/caio) is used internally which causes `AIOFile.write` to return an (awaitable) int (see [here](https://github.com/mosquito/aiofile/blob/8ff70ddc4e80ef8b127d1aad7d2b016c920d32dc/aiofile/aio.py#L220) and [here](https://github.com/mosquito/caio/blob/68acb69cf33b25c44c16f1b0366e3989f19db4f0/caio/asyncio_base.py#L108)). In aiofile<2.0.0, a different internal structure is used, and the return type of `AIOFile.write` is a byte string (which has a length, of course). 

**Fix**
I see two possible fixes:
1. Change the Pipfile to require aiofile<2.0.0.
2. Change the Pipfile to require aiofile>=2.0.0. Adapt to the new version of aiofile by changing [this line in async_file.py](https://github.com/bmoscon/cryptofeed/blob/da876616de5a2a76df2305a63549329f87335f45/cryptofeed/util/async_file.py#L32) from `self.pointer[uuid] += len(r)` to `self.pointer[uuid] += r`.

I'll open a PR for the second fix in a moment.

zanoo:
Implements the second fix from #282, and closes #282 

Summary:

- Fixes write bug in async_file.py
- Requires aiofile>=2.0.0 in Pipfile
- Formats Pipfile comments

EDIT: Also adds aiofile>=2.0.0 to setup.py

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,21 +16,21 @@ pandas = "*"
 requests = ">=2.18.4"
 sortedcontainers = ">=1.5.9"
 websockets = ">=7.0"
-"yapic.json" = ">=1.4.3"   # fast, drop-in replacement of the built-in python.json
+"yapic.json" = ">=1.4.3"    # fast, drop-in replacement of the built-in python.json
 # --- Optional packages to speed-up Cryptofeed -----------------------------------------------
-aiodns = ">=1.1"   # aiodns speeds up DNS resolving
-cchardet = "*"     # cchardet is a faster replacement for chardet
+aiodns = ">=1.1"            # aiodns speeds up DNS resolving
+cchardet = "*"              # cchardet is a faster replacement for chardet
 # --- Optional dependencies ------------------------------------------------------------------
-aiofile = ">=2.0.0"      # used by util/async_file.py
-aiokafka = "*"     # pip install cryptofeed[kafka]
-aioredis = "*"     # pip install cryptofeed[redis]
-arctic = "*"       # pip install cryptofeed[arctic]
-asyncpg = "*"      # pip install cryptofeed[postgres]
-motor = "*"        # pip install cryptofeed[mongo]
-aio_pika = "*"     # pip install cryptofeed[rabbit]
-pika = "*"         # used by demo_rabbitmq.py
-pyyaml = "*"       # pip install cryptofeed[rest_api]
-pyzmq = "*"        # pip install cryptofeed[zmq]
+aiofile = ">=2.0.0"         # used by util/async_file.py
+aiokafka = "*"              # pip install cryptofeed[kafka]
+aioredis = "*"              # pip install cryptofeed[redis]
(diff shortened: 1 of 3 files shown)


zanoo:
I just added the requirement to setup.py, thanks for pointing that out. No need to add myself to the contributors list for this.

PR #287, opened 2020-08-25
Changed binancefutures ticker to correct api endpoint

davidtwomey:
BinanceFutures now has its own `_address` and `message_handler` class functions (instead of using parent Binance).

NOTE: as discussed, some more work could be done in moving some code from Binance parent class which belongs solely to BinanceFutures (i.e. liquidations, open interest and funding).


bmoscon:
@davidtwomey thanks for the PR - please open a PR to add yourself to the authors file if you so desire!

PR #288, opened 2020-08-26
Influxdb auth username

akasimo:
I added username and password args for Influxdb authentication.

I tested:
- Authentication enabled server with username & password
- Authentication disabled server with username, password = None
Both are working ok.


bmoscon:
@akasimo thanks for the PR - let me test it and then I'll merge the PR

PR #289, opened 2020-08-28
Expose log filename and log level as parameters in FeedHandler and Rest initializers

zanoo:
This PR simply allows a user of the `Feedhandler` and `Rest` classes to specify the log filename and log level. Logging can then be done by the user like so:
```
import logging

f = Feedhandler(log_filename="log_filename_here.log", log_level=logging.INFO)
LOG = logging.getLogger("feedhandler")
LOG.info("starting feedhandler now")
f.run()
```
Existing filenames, levels, and log names are retained in default parameters.

NB: Setting the LOG object in feedhandler.py and rest.py is done using the `global` keyword; if you don't like this, let me know and I'll figure out something better.

### DIFF
--- a/cryptofeed/feedhandler.py
+++ b/cryptofeed/feedhandler.py
@@ -6,6 +6,7 @@ associated with this software.
 '''
 import asyncio
 import zlib
+import logging
 from collections import defaultdict
 from copy import deepcopy
 from socket import error as socket_error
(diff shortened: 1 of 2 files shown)


bmoscon:
I'm not opposed to this, just trying to think if an alternative is better - like reading it from an env var (and defaulting if not set), or doing like you have, but omit the global and just set it as a class attribute for use in feedhandler, or something else . . .

zanoo:
Yup, I agree. I like the environment variable + default approach. I'll implement it sometime this week and comment here when I make the commit.

zanoo:
Just updated it. Let me know what you think, and if the names of the env vars are ok.

PR #297, opened 2020-09-23
Support order_type in Trade Callback

zeppomedio:
Kraken (and perhaps other exchanges) sends down the `order_type` (market vs limit) that triggered a particular trade. This PR optionally passes this into the TRADES callback; to receive the order type, `include_order_type=True` must be passed when initializing the `TradeCallback` (to prevent a breaking change with previously written callbacks that didn't expect another arg or kwarg).

### DIFF
--- a/cryptofeed/callback.py
+++ b/cryptofeed/callback.py
@@ -25,8 +25,16 @@ class Callback:
 
 
 class TradeCallback(Callback):
-    async def __call__(self, *, feed: str, pair: str, side: str, amount: Decimal, price: Decimal, order_id=None, timestamp: float, receipt_timestamp: float):
-        await super().__call__(feed, pair, order_id, timestamp, side, amount, price, receipt_timestamp)
+
+    def __init__(self, callback, include_order_type=False):
+        self.include_order_type = include_order_type
+        super().__init__(callback)
+
+    async def __call__(self, *, feed: str, pair: str, side: str, amount: Decimal, price: Decimal, order_id=None, timestamp: float, receipt_timestamp: float, order_type: str=None):
+        kwargs = {}
+        if self.include_order_type:
+            kwargs['order_type'] = order_type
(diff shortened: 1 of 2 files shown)


cryptofeed/callback.py
```diff
@@ -25,8 +25,16 @@ async def __call__(self, *args, **kwargs):
 
 
 class TradeCallback(Callback):
-    async def __call__(self, *, feed: str, pair: str, side: str, amount: Decimal, price: Decimal, order_id=None, timestamp: float, receipt_timestamp: float):
-        await super().__call__(feed, pair, order_id, timestamp, side, amount, price, receipt_timestamp)
+
+    def __init__(self, callback, include_order_type=False):
+        self.include_order_type = include_order_type
```
bmoscon, line 30: I see, rather than have it enabled by default, its user configured. Interesting, I wonder if it would be advantageous to have it always on vs user enabled. Makes me also wonder about other fields, maybe they should all be configurable. Things to think about - I think this is good for now, I'll think about the ramifications of some of those other thoughts and let you know if I'm leaning, in the future, to something that deviates from this significantly. Only other thought is to have an exclude list, thats normally empty for all CBs but for trades the default is include order type, which then would require the user to un-exclude it. idk. I tend to come up with my best ideas when I'm away from the computer so will think about this.

bmoscon (approved):
Looks good - thanks!

bmoscon:
@zeppomedio see my solitary comment - no changes needed - just some random thoughts about how this may (but may not) change in the future. appreciate the PR

cryptofeed/callback.py
zeppomedio: I think for simplicity always-on would be best, but I was wary of breaking already-defined callbacks out in the wild. If there ends up being more fields to add in the future, perhaps wrapping all of the args in a `Trade` object or similar would simplify things so that new fields can be added without breaking the function signature (other than the one-time break to go towards a single arg rather than the list)

PR #298, opened 2020-09-24
Optionally keep track of order type for Coinbase, if L3_BOOK channel is enabled

zeppomedio:
Building off of #297, this adds support for order type (market/limit) to the taking side of a trade for Coinbase. 

Totally open to keeping a branch up to date instead of merging if this seems like too much intervention for this feature. I don't love that we now check for `keep_order_book` in multiple places, very open to different suggestions.

Test plan:
1) Subscribe to Coinbase with:
```
fh.add_feed(
    Coinbase(pairs=["BTC-USD"], channels=[L3_BOOK, TRADES], callbacks={
        TRADES: TradeCallback(process_trade, include_order_type=True)
    }
))
```
Verify that Order Type is sent to the process_trade callback

2) Subscribe to Coinbase with:
```
fh.add_feed(
    Coinbase(pairs=["BTC-USD"], channels=[TRADES], callbacks={
        TRADES: TradeCallback(process_trade, include_order_type=True)
    }
))
```
Verify that order_type=None is sent to Callback, since we haven't subscribed to the L3_BOOK.

3) Subscribe to Coinbase with:
```
fh.add_feed(
    Coinbase(pairs=["BTC-USD"], channels=[L3_BOOK, TRADES], callbacks={
        TRADES: TradeCallback(process_trade, include_order_type=True),
        L3_BOOK: BookCallback(process_book)
    }
))
```

Verify that `process_book` is correctly called (in other words verify that `keep_order_book` is True because we have at least one book callback, and the book is being correctly kept).

### DIFF
--- a/cryptofeed/exchange/coinbase.py
+++ b/cryptofeed/exchange/coinbase.py
@@ -13,7 +13,7 @@ import requests
 from sortedcontainers import SortedDict as sd
 from yapic import json
 
-from cryptofeed.defines import BID, ASK, BUY, COINBASE, L2_BOOK, L3_BOOK, SELL, TICKER, TRADES
+from cryptofeed.defines import BID, ASK, BOOK_DELTA, BUY, COINBASE, L2_BOOK, L3_BOOK, SELL, TICKER, TRADES
 from cryptofeed.feed import Feed
 from cryptofeed.standards import pair_exchange_to_std, timestamp_normalize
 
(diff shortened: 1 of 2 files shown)


cryptofeed/exchange/coinbase.py
```diff
@@ -26,10 +26,16 @@ class Coinbase(Feed):
 
     def __init__(self, pairs=None, channels=None, callbacks=None, **kwargs):
         super().__init__('wss://ws-feed.pro.coinbase.com', pairs=pairs, channels=channels, callbacks=callbacks, **kwargs)
+        # we only keep track of the order book if we have at least one subscribed order-book callback
+        self.keep_order_book = False
+        for cb_type in callbacks:
+            if cb_type in (L3_BOOK, L2_BOOK, BOOK_DELTA):
```
bmoscon, line 32: Pretty sure L2 book doesnt need to be touched, does it? You'd only ignore L3 books and L3 deltas if L3 was enabled but there was not callback for it

bmoscon (changes requested):

bmoscon:
@zeppomedio looks good other than the fact I don't think L2 book need to be impacted by this change

bmoscon (approved):

bmoscon:
@zeppomedio i think this is fine, optionally you could just ignore the l3 book events in the message handler, might be easier or cleaner to you to do it that way. You would have to parse out the symbol/trading pair before hand though and that may be cumbersome. idk, i'd have to think about it, but this way doesn't seem bad. if you're ready for this to be merged lmk

zeppomedio:
Thanks @bmoscon, good to go on my end.

PR #319, opened 2020-11-02
New Exchange: Probit

qinghuangchan:
### Description of code 

- [x] - Tested:
- output to console 
- [x] - Changelog updated

Support for Probit wss://api.probit.com/api/exchange/v1/ws endpoint

Added support for the `TRADES` and `L2_BOOK`. 

Probit has a "ticker" channel but provides OHLC data, not BBO, so i've left that out. 

To call a pair, use 'usual format' {BTC-USD, ETH-USD, ETH-BTC}


cryptofeed/exchange/probit.py
```diff
@@ -0,0 +1,191 @@
...
+                                                     }))
+        else:
+            for pair in self.pairs:
+                await websocket.send(json.dumps({"type": "subscribe",
+                                                 "channel": "marketdata",
+                                                 "filter": list(self.channels),
+                                                 "interval": 100,
+                                                 "market_id": pair,
+                                                 }))
+                await asyncio.sleep(0.3)
```
bmoscon, line 191: is this needed? Unless you need to sync data from a REST endpoint or the like I'm guessing this is not needed

bmoscon (changes requested):
feedhandler.py also has a dict of string to object mappings that is used, can you add PROBIT here as well?

cryptofeed/exchange/probit.py
qinghuangchan: you're right, the `await asyncio.sleep(0.3)` is probably not needed. Removing.

bmoscon (approved):

PR #322, opened 2020-11-04
Use more clear rate limits values for FTX fundings

jerry40:
### Currently FTX fundings rate limit can vary, it depends on pairs count. If you subscribe on 1 pair, the rate is 1/30 seconds which is too much for fundings I think. I propose sending 10 requests per second and wait 60 seconds between loops (perhaps it is too long?)

- [ yes ] - Tested
- [ yes ] - Changelog updated
- [ no ] - Contributors file updated (optional)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@
   * Feature: Arctic library quota can be configured, new default is unlimited
   * Feature: New exchange: Probit
   * Bugfix: Correctly store receipt timestamp in mongo backend
+  * Change: FTX - set a funding rate requests limit constant (10 requests per second, 60 seconds pause between loops)
   
 ### 1.6.0 (2020-09-28)
   * Feature: Validate FTX book checksums (optionally enabled)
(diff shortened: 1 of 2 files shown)


jerry40:
Or should we just change `wait_time = len(pairs) / 30` to `wait_time = 30. / len(pairs)` to be able to refresh all rates every 30 seconds?

bmoscon:
Is there an issue where you are getting rate limited? I'm not sure that potentially waiting 60 seconds for an update is a good idea either.

jerry40:
Currently it is about 62 perpetuals on FTX, so I get a funding every 2 seconds, so to have all fundings I need to wait for 2 minutes which is a bit longer than I can expect. A 2 minute gap between two loop starts is even bigger than 60 seconds I proposed.
Also, if I use only one pair, it starts fetching fundings every 1/30 sec which is an overkill for a value that refreshes once a several hours. Let's at least use `wait_time = 30. / len(pairs)` then? So every 30 seconds you have full actual funding data.

bmoscon (approved):

bmoscon:
thanks for the PR!

PR #337, opened 2020-11-27
Add stop callback on shutdown

qinghuangchan:
In this PR, I propose to add some additional handling when exiting. 

Background for needing this feature:

I've opted to store my data in pgsql. The current pgsql insertion implementation doesn't do any caching, so i've monkey patched my own caching methods in.

However, when restarting the collector, the data in cache is lost, since there is no opportunity to do clean up. 

I've added a `stop` method to the `Callback` class, as well as created a new `BackendCallback` base class which all other backend callbacks inherit. 

The purpose of adding such handling is to allow the many types of callbacks to do some clean up where necessary. 

In the `run` method of `FeedHandler`, I've added a `finally` clause to the try-except statement. This is where the code will try to call `stop` for all its feeds. 

Im not too sure if you'd like my implementation - please let me know if there's something I can change to make it better. 

- [x] - Tested
- [x] - Changelog updated

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,10 @@
 ## Changelog
-  * Feature: Add stop callback when exiting loop, add stop method placeholder for base callbacks
 
 ### 1.6.2
   * Bugfix: Fix aggregated callbacks (Renko and OHLCV) when used with exchanges that support order types
   * Bugfix: Fix broken example/demo code
   * Feature: New data channel - `futures_index` - demonstrated in ByBit
+  * Feature: Add stop callback when exiting loop, add stop method placeholder for base callbacks
 
 ### 1.6.1 (2020-11-12)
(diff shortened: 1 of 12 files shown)


CHANGES.md
```diff
@@ -1,4 +1,5 @@
 ## Changelog
+  * Feature: Add stop callback when exiting loop, add stop method placeholder for base callbacks
```
bmoscon, line 2: move this under the current dev version, 1.6.2

bmoscon (changes requested):
rather than put in an un-used stub, why not just check if stop is a defined method on `callback` before calling it? This would remove the need for all the added in inheritance

qinghuangchan:
That's looks nicer indeed. changes committed.

CHANGES.md
qinghuangchan: done.

bmoscon (approved):

PR #339, opened 2020-11-28
Add optional caching mechanism to postgres backend

qinghuangchan:
In this PR, I propose to add a cache for the `PostgresCallback` backend. Without it, asyncpg spends way too much time handling connection overheads. 

I've set the default behavior to NOT cache so that new users will be able to see the data in their DB immediately and not have to wonder why its not appearing. More serious users should be able to find the cache param easily. 

On shutdown, the `stop` method will write the existing cache into DB, so nothing will be lost. 

Please let me know if there's something I can change to make this PR better.

- [x] - Tested
- [x] - Changelog updated

### DIFF
--- a/cryptofeed/backends/postgres.py
+++ b/cryptofeed/backends/postgres.py
@@ -5,7 +5,6 @@ Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
 from datetime import datetime as dt
-import asyncio
 
 import asyncpg
 from yapic import json
@@ -43,7 +42,6 @@ class PostgresCallback:
         self._cache_size = cache_size
         self._cache = []
         self._cache_counter = 0
-        self._cache_lock = asyncio.Lock()
 
     async def _connect(self):
         if self.conn is None:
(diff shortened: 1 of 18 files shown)


cryptofeed/backends/postgres.py
```diff
@@ -37,17 +40,44 @@ def __init__(self, host='127.0.0.1', user=None, pw=None, db=None, table=None, nu
...
+
     async def _connect(self):
         if self.conn is None:
             self.conn = await asyncpg.connect(user=self.user, password=self.pw, database=self.db, host=self.host)
 
     async def write(self, feed: str, pair: str, timestamp: float, receipt_timestamp: float, data: dict):
+        time = dt.utcfromtimestamp(timestamp)
+        rtime = dt.utcfromtimestamp(receipt_timestamp)
+
+        async with self._cache_lock:
```
bmoscon, line 56: why do you need a lock? There are no awaits in this code block

bmoscon: also, you can easily avoid the locking in the case where the cache is not used, so at a minimum the locking should only be in the case where the cache is being utilized. I'm still not convinced its needed for that case though.

qinghuangchan: You are right, there is no need for a lock. I had a previous implementation where it was needed, but didnt think to remove the lock with this current implementation. Changes comitted.

bmoscon (approved):

PR #344, opened 2020-12-04
Fix simple feedhandler example

vakili:
Thank you for this project!

I noticed that the feedhandler example in `docs/high_level.md` produces errors due to defining callback functions with the wrong number of arguments.

I simply supplied the correct number of arguments, and updated one print statement to include a timestamp.

- [x] - Tested (by running the example)
- [ ] - Changelog updated


bmoscon (approved):

bmoscon:
thanks for the contribution!

PR #348, opened 2020-12-06
Add GCP Pub/Sub backend

globophobe:
Add GCP Pub/Sub backend

- [x] - Tested with GCP Pub/Sub in the cloud, and locally with emulator 
- [X] - Changelog updated
- [X] - Contributors file updated (optional)

### DIFF
--- a/cryptofeed/backends/gcppubsub.py
+++ b/cryptofeed/backends/gcppubsub.py
@@ -1,0 +1,123 @@
+'''
+Copyright (C) 2017-2020  Bryant Moscon - bmoscon@gmail.com
+
+Please see the LICENSE file for the terms and conditions
+associated with this software.
+'''
+import os
+import io
+from typing import Optional
+from typing import IO
+from typing import Union
+from typing import AnyStr
+
+import aiohttp
+import google.api_core.exceptions
+from google.cloud import pubsub_v1
+from yapic import json
+
+# Use gcloud.aio.pubsub for asyncio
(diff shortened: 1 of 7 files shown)


bmoscon (changes requested):
could you provide a bit of documentation, either in the code (in comments) or in a file in the `docs/` dir that explains how you'd use this in a "real" setting? Presumably keys/endpoints/etc need to be set up in various env vars, etc. I'm sure people could figure it out via the code, but its easier if its just in a block comment at the top or in a doc.

cryptofeed/backends/gcppubsub.py
```diff
@@ -0,0 +1,94 @@
+'''
+Copyright (C) 2017-2020  Bryant Moscon - bmoscon@gmail.com
+
+Please see the LICENSE file for the terms and conditions
+associated with this software.
+'''
+import os
+
+import aiohttp
+import google.api_core.exceptions
```
bmoscon, line 10: these google imports will need to be added to the `extras` in setup.py. You can list them under something like 'gcp' or 'GCPPubSub' if you like

bmoscon (changes requested):

cryptofeed/backends/gcppubsub.py
globophobe: Added as 'gcp_pubsub'

bmoscon (approved):

PR #366, opened 2020-12-23
Refactor FeedHandler websocket

micheartin:
* Add class `WebsocketFeed` because `RestFeed` has different parameters for `subscribe()`
* Use an adaptive timeout algo to detect sooner a ghost websocket connection (with less false positive)
* Prepare support for OKEX options (full PR will be sent later)

### DIFF
--- a/cryptofeed/exchange/okcoin.py
+++ b/cryptofeed/exchange/okcoin.py
@@ -9,9 +9,11 @@ from itertools import islice
 import logging
 import zlib
 
+import websockets
 from sortedcontainers import SortedDict as sd
 from yapic import json
 
+from cryptofeed import feed
 from cryptofeed.defines import ASK, BID, BUY, FUNDING, L2_BOOK, OKCOIN, OPEN_INTEREST, SELL, TICKER, TRADES, LIQUIDATIONS
 from cryptofeed.exceptions import BadChecksum
 from cryptofeed.feed import Feed
@@ -21,7 +23,7 @@ from cryptofeed.standards import pair_exchange_to_std, timestamp_normalize
 LOG = logging.getLogger('feedhandler')
 
 
-class OKCoin(Feed):
+class OKCoin(feed.WebsocketFeed):
     id = OKCOIN
 
     def __init__(self, pairs=None, channels=None, callbacks=None, **kwargs):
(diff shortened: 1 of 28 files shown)


cryptofeed/exchange/probit.py
```diff
@@ -165,7 +166,7 @@ async def message_handler(self, msg: str, timestamp: float):
             await self._l2_update(msg, timestamp)
         # Probit has a 'ticker' channel, but it provide OHLC-last data, not BBO px.
 
-    async def subscribe(self, websocket):
+    async def subscribe(self, websocket: websockets.WebSocketClientProtocol) -> bool:
```
bmoscon, line 169: why are the subscribe methods all returning a bool?

olibre: Because some exchanges do not support all channels as websocket subscription.
Therefore some WebsocketFeed poll REST endpoints to retrieve data for some channels.
I am searching an example...

cryptofeed/exchange/okcoin.py
```diff
@@ -54,32 +56,60 @@ def __calc_checksum(self, pair):
...
+
+        for args in smaller_lists:
+            LOG.info("%s: Subscribe to %s args from %r to %r", self.id, len(args), args[0], args[-1])
+            request = {"op": "subscribe", "args": args}
+            await websocket.send(json.dumps(request))
+            # wait for response (max 1 second) before next subscription batch
+            msg = await asyncio.wait_for(websocket.recv(), timeout=1)
+            if msg:
+                await self.message_handler(msg, time.time_ns())
+        return True
```
bmoscon, line 112: this code in subscribe seems interesting and useful. I'm not sure of the problem its solving, but would be nice to get this in.w

olibre: This code avoids flooding exchange with too many subscribe requests: we wait at least one response with one second timeout. And this code avoids buffering received responses.

bmoscon: looks pretty good, do you want to port this change over individually? I'm happy to copy/paste it and then test it and get it merged, but its your code and contribution so I'm fine waiting for you to do it if you want to.

olibre: :ok: :+1: I am preparing a PR to propose a refactoring/cleaning your new design using `AsyncConnection`

bmoscon: :+1:

olibre: Happy New Year Bryant @bmoscon 

I have worked hard to improve your new `AsyncConnection` design and I will propose you a nice PR with a new file `runner.py` implementing the class `Runner`. The `Runner` owns the main `message_handler` infinite loop ( and `_watch()`) that was previously the responsibility of `FeedHandler`.

In the coming PR, I have replaced the `Feed.connections()` function by reusing the `address: dict` in the `Feed.__init__()` (first argument). No need to pass `partial(subscribe,x)`
 and `message_handler()` callbacks. The `address: dict` (options and URLs) are used by the `FeedHandler` to create the `AsyncConnection`: each `AsyncConnection` keeps its option/address within its context. This connection context is then accessed by the `Feed` subclass to subscribe to the right pairs/channels.

The `FeedHandler` manages the following objects:
* `Feed` (exchange or provider)
* `AsyncConnection` (Websocket or HTTP)
* `Runner` (one `Runner` per `AsyncConnection`)

The `FeedHandler` has the responsibility to create the `Feed` subsclass, to create its the corresponding `AsyncConnection` according to the `Feed.address` (`str` or `dict`, can be a mix of WS and HTTP), and to create one `Runner` per `AsyncConnection`. Each `Feed` has one or more `AsyncConnection`. Each `Runner` runs only one `AsyncConnection`.

`Runner` knows/calls `Feed` and `AsyncConnection`. But the `Feed` and its subclasses do not know the `Runner`. The `Runner` pass the `AsyncConnection` when calling `Feed.subscribe() and `Feed.handle()`. The `Feed` subclasses use the `AsyncConnection` to access the connection context and to send messages (websocket only). Notice I have renamed `message_handler()` -> `handle()` for consistency with `subscribe() != message_subscriber()`.

Once everything is created, the `FeedHandler` runs all the `Runner` asynchronously. The `Runner` infinite loop calls the `Feed.subscribe(conn)`, and loops over the `AsyncConnection.read()`. Each received message is passed to `Feed.handle(msg,timestamp,conn)`. The `Feed` subcalss has access to `conn.ctx`, the connection context: no need to mange a `seqno_map[conn.id]`. The `conn.ctx` contains the `Feed.address` (and options) used to create the `AsyncConnection`.

All exchanges using websocket benefits this feature:
```py
# slow down the subscription batch by waiting for a response or for a half of a second before next send
try:
    msg = await asyncio.wait_for(self.socket.recv(), timeout=0.5)
    await self.handle(msg, time(), self)
except asyncio.exceptions.TimeoutError:
    pass
```

On shutdown signal, the `FeedHandler` closes the `AsyncConnection`, stops the `Runner` & `Feed`, and waits for termination of all backends before exiting. The `FeedHandler` runs this functions using AsyncIO to speed up the shutdown process. This is important because multiple backends may need to flush their data and the orchestrator (K8s/Docker) may kill Cryptofeed is it takes too much time to terminate. However, even if the backend flushing is faster, `loop.run_until_complete(asyncio.gather(*flush_tasks))` takes some seconds to detect all flushing tasks are completed.

I have adapted almost all exchanges and providers: I have to finish WHALES_ALERT. This improvement of your new design simplifies the source code, especially BITFINEX, BINANCE and more... I have also tested long run with all exchanges/providers activated (except WHALES_ALERT), subscribing to many many pairs per `Feed`, and with backends requiring final flush.

PR will be ready at the end of the week...

bmoscon: @olibre  bonne année to you as well. Looking forward to it!

micheartin:
Replaced by https://github.com/bmoscon/cryptofeed/pull/381

PR #367, opened 2020-12-24
Connection refactor

bmoscon:
* New connection design. Handles rest and websocket connections seamlessly for websocket and REST endpoint exchanges. Will allow exchanges to use more than 1 websocket connection seamlessly. 
* Update BitMax to use their new pro API

- [X] - Tested
- [X] - Tests run and pass


bmoscon:
@yohplala just a heads up - I reworked coingeko. Coingeko now takes a symbol (like BTC) and no longer a pair (like BTC_USD) since the api only works with the base and not the quote, it seemed weird and misleading to have the quote information in there. `pairs` will probably need to be renamed project wide to either symbol(s) or instrument(s) to be more consistent as we have more and more things that are no longer simple trading pairs. Also, I did NOT rework Whale Alert since I do not have an API key and I haven't begun completed my work on the authenticated exchanges yet (this is step 1 for that). I can fix it eventually, but you might want to take a look at the changes I made to CoinGecko and use that as a template for fixing it yourself.

bmoscon:
@olibre this is the connection refactor I was mentioning earlier. I've tested it for both websocket and rest feeds. I think your PR has a lot of good things, but it will need to have the merge conflicts resolved after this goes in. Feel free to take a look and leave a review

yohplala:
> @yohplala just a heads up - I reworked coingeko. Coingeko now takes a symbol (like BTC) and no longer a pair (like BTC_USD) since the api only works with the base and not the quote, it seemed weird and misleading to have the quote information in there. `pairs` will probably need to be renamed project wide to either symbol(s) or instrument(s) to be more consistent as we have more and more things that are no longer simple trading pairs.

Hi @bmoscon,
Thanks for the message.
 - Coingecko / use of `symbol` vs `pair`:  a `pair` is asked for initially because some market info are given using a `quote_currency`, like `price` (of couse), but also `market_capitalization` and a few others. Coingecko is providing a loooooot of `quote_currencies` and I restricted choice for user only to `USD`, `BTC`, and `ETH`. We can further restrict/force it to `USD` and this way not provide to the user the ability to select a `quote_currency`, I am fine with that.
I had a look at what you did in `coingecko.py`, and I can see you replaced in my code both `quote_c` and `base_c` with `pair`.
It will work without error message, but then, considering BTC for instance, it means price of BTC will be in BTC. Basically, result will always be 1 :). And market cap will be equal to its circulating supply. (I asked to Coingecko, their market cap is obtained by multiplying price to circulating supply, and not total supply)
=> so there is a minor work here. I can have a look this evening. Just tell me what you are fine with:
- either we force `quote_currency` to `USD`  and only a `symbol`is indeed needed
- or we leave a `pair` to the user

yohplala:
> Also, I did NOT rework Whale Alert since I do not have an API key and I haven't begun completed my work on the authenticated exchanges yet (this is step 1 for that). I can fix it eventually, but you might want to take a look at the changes I made to CoinGecko and use that as a template for fixing it yourself.

Ok, I can start have a look this evening as well. Not sure I can do it in one evening, so probably it will last till end of this week-end.
FYI, you can have from Whale Alert a free license by using the free plan. It is available on their website.

yohplala:
> > Also, I did NOT rework Whale Alert since I do not have an API key and I haven't begun completed my work on the authenticated exchanges yet (this is step 1 for that). I can fix it eventually, but you might want to take a look at the changes I made to CoinGecko and use that as a template for fixing it yourself.
> 
> Ok, I can start have a look this evening as well. Not sure I can do it in one evening, so probably it will last till end of this week-end.
> FYI, you can have from Whale Alert a free license by using the free plan. It is available on their website.
> This is what I used during the development of the feed.

Whoooooops, I am wrong, only the 1st month is free. I did not recall that.

bmoscon:
ok, I think I fixed coingecko - the user can still specify a normal symbol (btc-usd) and the data parsing will now include the correct currency info. The caveat is that the pair generation function will not show btc-usd as a legal value, since its not. personally i think it should be a hybrid type approach where you can do either or, if you follow the guidelines from the exchange and only use btc, you get all the metrics in all the currencies, if you specify a quote currency, you get just that one. Sometime I can look into later

olibre:
Hi @bmoscon and @yohplala 
I will be absent for a couple of days, my apologies for my lack of contribution these days/weeks/months.

On my side I have hacked many parts of Cryptofeed: implemented another Whale-Alert and CoinGecko retrievals (created the channel SCORE that is different from MARKET_INFO). The subscription is also based on currencies, for Whale-Alert, CoinGecko and Bitfinex. On Bitfinex, I have fixed the Trade Funding and added the Ticker Funding (new channel TICKER_FUNDING). I have also added support for JSON library [orjson](https://pypi.org/project/orjson/) and built-in json lib. Also use of `uvloop` that is about 4 times faster than built-in `asyncio`. The receipt and market timestamps are in nanoseconds `time.time_ns()` because nanosecond is the new standard for timestamps and `float` cannot store the nanosecond precision (nanoseconds are back converted to seconds when required: in many backends and demo samples). OKEx source code also supports options.

The changes are huge, and I will spend time to send multiple Pull Requests. But maybe, the simplest for me is to push everything and to use it as a new base code for Cryptofeed. I took care about all backend source code, but I not not have tested all backends, so I do not know some thing is broken. I have also successfully ran Cryptofeed with Dockerfiles based on Clear Linux (the fastest Pandas/Numpy binaries) and also on another Dockerfile based on PyPy.

Please let me know how you want to proceed. It is faster to hack as a monkey, changing many parts everywhere. But I also need to take some time to share my improvements to the awesome upstream Cryptofeed project and community. :heart:

bmoscon:
@olibre orjson doesnt support parsing into Decimal from float (unless that has changed recently) and I've profiled my code with uvloop and plain asyncio and I don't see any performance improvement. I'm open to orjson, as long as it parses floats to Decimal, and I'd only be interested in uvloop provided that there were real metrics that show and improvement.

yohplala:
> The caveat is that the pair generation function will not show btc-usd as a legal value, since its not

@bmoscon , yes, it is actually possible to check it is a legal value.
From `BTC-USD` given by the user, we should split this input in `base_currency` and `quote_currency`.
As already done by existing code, `base_currency` is to be checked using end point:
`https://api.coingecko.com/api/v3/coins/list`

`quote_currency`is then to be checked using endpoint
`https://api.coingecko.com/api/v3/simple/supported_vs_currencies`

Initially, I coded something along these lines, except I generated the full list of accepted pairs, meaning generation of ~ 1500 `base_currency` x ?30 `quote_currency`... Well far too much CPU / memory consumption for the relevance of this function...
Now I see we should just keep both lists separated, split the pair given by the user, and check them independently.

Bryant, I am sorry, it's Christmas Eve, some wine glass have gone by and I don't feel the code I could produce this evening will generate anything except error messages... I am dropping my turn this evening, and probably tomorrow as well, sorry :)

yohplala:
> if you follow the guidelines from the exchange and only use btc, you get all the metrics in all the currencies, if you specify a quote currency, you get just that one.

Ths is a very relevant approach in my opinion.

bmoscon:
@olibre i added in support for uvloop. From what I've profiled, it doesn't hurt anything, but I'm not sure I've seen any benefit either. It's a tiny change, so no big deal. I am curious how you saw an improvement though, so please let me know at your convenience.

olibre:
Hi @bmoscon Merry xmas

Sorry for the confusion: I have not profiled Cryptofeed with uvloop.
The 4x improvements is just the advertisement from uvloop project:  
https://github.com/MagicStack/uvloop#performance
The global gain for Cryptofeed could be about a couple of percentages...

I am successfully using orjson without `float` issues: I have removed all `float` -> `Decimal` conversions in exchanges source code. When the `float` is then converted to string, we obtain the same value. Moreover, when the prices (and amounts) are expressed in a string, I do not perform any string -> number conversion because most of the backends are happy passing the value as a string. 

BUT (there is a big BUT), the order-book functions require to manipulate prices/amount as number. I have not yet checked all configurations (L2 and/or L3 with all exchanges / backends).

Are you still interested to support orjson and standard built-in json libraries along with yapic.json?

bmoscon:
Ok, I know orjson is faster than yapic. Yapic is faster than the other options, but orjson is def faster. Because the orderbooks are sorted, the keys have to be decimal.Decimal objects. strings will not work. For consistency I think its ideal to just have all numerics as Decimal. you are right, from float -> decimal, we dont lose/gain anything, but from string -> decimal we can ensure at least we are not causing any precision loss. If you want to make the change and open a PR, I'll take it, otherwise I can take a crack at it. It shouldn't be a huge change; lots of files changed, yes, but not lots of lines

bmoscon:
@olibre - I've been doing a lot of code profiling to find pain points that can be improved. I did some testing with ORjson vs yapic.json, and I'm not convinced that orjson will benefit this at all. from several test runs looking at 100k+ parsed messages ORJson averages about 8.5 microseconds per message (for json.loads). yapic averages 6.8 microseconds for a similar volume. This also doesnt even take into account that yapic also parses the datetime for as well, saving time later when the datetime string will need to be converted to a UTC timestamp. If other profiling shows improvements, I'm all ears, but based on what I've seen today while profiling against coinbase, I don't think orjson will give any benefits. Maybe in general it is better or faster, but I'm more persuaded by the profiling results with real exchange data

olibre:
Dear @bmoscon 
You are right to profiling the different options before deciding. :+1: 

In my refactored Cryptofeed, I have removed the price (and amount) conversion from `string` to `Decimal` and pass the `string` value as it to the backends, except for L3 and L2 books.

I am also running Cryptofeed using PyPy 3.7 but ORJson and yapic do not support PyPy. Therefore I have implemented an import try/catch in Cryptofeed to select by priority: yapic, else ORJson, else built-in JSON lib.

I have to send you some PRs but my priorities have changed, and I do not know when I will have time to put again my hands in Cryptofeed source code…

Issue #371, opened 2020-12-27
Pipenv installation broken


UnitNote:
**Describe the bug**
When installing via `git clone` and `pipenv`, the user must also run `python3 -m pip install .`, running `python3 -m pipenv install` is not enough (with Python 3.7.5), but the INSTALL.md documention does not mention this.

The error appears to be that the pipenv install does not result in uvloop (or cryptofeed itself ) being installed.

**To Reproduce**
```
user@Ubuntu18:~/cryptofeed$ git log | head -n 1
commit 6eafd56b98a6a895d7485d46841e908992082e92

user@Ubuntu18:~/cryptofeed$ python3.7 -m pipenv install
Creating a virtualenv for this project...
Pipfile: /home/user/cryptofeed/Pipfile
Using /usr/bin/python3.7 (3.7.5) to create virtualenv...
⠇ Creating virtual environment...created virtual environment CPython3.7.5.final.0-64 in 433ms
  creator CPython3Posix(dest=/home/user/.local/share/virtualenvs/cryptofeed-XkEPfgbK, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/user/.local/share/virtualenv)
    added seed packages: pip==20.3.1, setuptools==51.0.0, wheel==0.36.1
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator

✔ Successfully created virtual environment! 
Virtualenv location: /home/user/.local/share/virtualenvs/cryptofeed-XkEPfgbK
Installing dependencies from Pipfile.lock (ba6a78)...
Ignoring idna-ssl: markers 'python_version < "3.7"' don't match your environment
Ignoring typing: markers 'python_version < "3.7"' don't match your environment
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 64/64 — 00:00:54
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.

user@Ubuntu18:~/cryptofeed$ pipenv shell
Launching subshell in virtual environment...
user@Ubuntu18:~/cryptofeed$  . /home/user/.local/share/virtualenvs/cryptofeed-XkEPfgbK/bin/activate

[...]

PR #378, opened 2021-01-06
Explain when Cryptofeed crashes during pairs retrieval

olibre:
During the pairs retrieval at startup time, some exchanges may reply an erroneous response.
Currently Cryptofeed does not always handle the exception, and crashes.

Some users think this is a bug in Cryptofeed: https://github.com/bmoscon/cryptofeed/issues/371#issuecomment-751899389

This PR improves the user experience by providing a clear explanation of the failure.

### DIFF
--- a/cryptofeed/pairs.py
+++ b/cryptofeed/pairs.py
@@ -9,10 +9,12 @@ Pair generation code for exchanges
 '''
 
 import logging
+import time
 from collections import defaultdict
-from typing import Dict, List, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import requests
+from requests import Response
 
 from cryptofeed.defines import *
 
@@ -21,8 +23,20 @@ LOG = logging.getLogger('feedhandler')
 
 PAIR_SEP = '-'
 
-_pairs_retrieval_cache = dict()
+_pairs_retrieval_cache: Dict[str, Dict[str, str]] = {}
 _exchange_info = defaultdict(lambda: defaultdict(dict))
+
+
+def raise_failure_explanation(feed_id: str, exception: BaseException, responses: Dict[str, Optional[Response]]):
+    LOG.critical('%s: encountered %r while processing response from exchange API', feed_id, exception)
+    if len(responses) > 1:
+        LOG.critical('%s: content of one of the %s responses was unexpected', feed_id, len(responses))
+    for url, r in responses.items():
+        if url:
+            LOG.critical('%s: requested URL: %s', feed_id, url)
+        if r:
+            LOG.critical('%s: unexpected response: %s', feed_id, r.text)
+    raise ValueError(f'Cryptofeed stopped because of an unexpected response from {feed_id}') from exception
 
 
 def set_pair_separator(symbol: str):
@@ -53,9 +67,7 @@ def _binance_pairs(endpoint: str, exchange: str):
                 _exchange_info[exchange]['contract_type'] = symbol['contractType']
         return ret
     except Exception as why:
-        LOG.critical('%s: encountered %r while processing exchangeInfo response from exchange API', exchange, why)
-        LOG.critical('%s: unexpected response: %r', exchange, r.text if r else r)
-        raise ValueError(f'Cryptofeed stopped because of an unexpected response from {exchange}') from why
+        raise_failure_explanation(exchange, why, {endpoint: r})
 
 
 def binance_pairs() -> Dict[str, str]:
@@ -91,7 +103,8 @@ def bitfinex_pairs() -> Dict[str, str]:
         ret = {}
         for pair in [t[0] for t in tickers]:
             if pair[0] == 'f':
-                pass  # normalized = norm.get(pair[1:], pair[1:])
+                # We will repair FUNDING soon on Binance, and enable the following line
+                continue # normalized = norm.get(pair[1:], pair[1:])
             else:
                 if len(pair) == 7:
                     base, quote = pair[1:4], pair[4:]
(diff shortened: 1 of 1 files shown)


cryptofeed/pairs.py
```diff
@@ -324,59 +431,117 @@ def bitmax_pairs() -> Dict[str, str]:
...
-    data = requests.get('https://api.coingecko.com/api/v3/coins/list').json()
-    return {entry['symbol'].upper(): entry['id'] for entry in data}
+    r = None
+    try:
+        r = requests.get('https://api.coingecko.com/api/v3/coins/list')
+        return {entry['symbol'].upper(): entry['id'] for entry in r.json()}
+    except Exception as why:
+        LOG.critical('COINGECKO: encountered %r while processing response from exchange API', why)
+        LOG.critical('COINGECKO: unexpected response: %r', r.text if r else r)
+        raise ValueError('Cryptofeed stopped because of an unexpected response from COINGECKO') from why
```
bmoscon, line 532: since nearly all the error messages are the same format - exchange something something, could we just store the strings as global/consts and just format them in place in all the log lines? Then if ever the format needs to be slightly tweaked or adjusted, there is only one place to do it?

bmoscon (changes requested):

bmoscon:
looks good other than this one change. I can make the change myself it you do not have the time, just let me know

cryptofeed/pairs.py
olibre: Good idea :+1:
As you prefer, I can pool the strings myself or let you do it...
Please let me know :wink:

olibre: Hey @bmoscon I am working on this requested improvement... :rocket:

olibre:
Hi Bryant
I have added a function to log and raise the shared failure explanation.
There is also a commit to normalize IOTA in CoinGecko.
I wish you the best for 2021:sparkles:

bmoscon (approved):

bmoscon:
thanks for the PR!

olibre:
Hi Bryant 
I have introduced a regression in that PR. Do not publish a new release, I will send you a fix.

The issue with CoinGecko is the huge number of supported pairs, and the difficulty to list all of them. The endpoint https://api.coingecko.com/api/v3/coins/list returns 5441 quote assets. I think there are about 15400 combinations of pairs `{BaseAsset}-{QuoteAsset}`. I will try to find a way to provide them. There is about 232 bases, 70 quotes and 2430 assets. The total of unique active assets is 2503.

yohplala:
Hi @olibre ,

I thought about that and had in mind to add to the dict of base asset a last `_QuoteAsset` key that would be the list of the quote currencies.

Then, in the function checking that pair exist,

- either it proceeds as currently coded (checking existnce of a pair in pair_dict)
- or, within a new if to implement:
       if `_QuoteAsset`key exist in pair dict:
              then,
                    # checks are separate
                    check that `BaseAsset` exists in key
                    check that `QuoteAsset` exists in pair_dict['_QuoteAsset`

bmoscon:
the pairs functions need to return all legal symbols/pairs/etc, so however you want to do it, thats fine by me

olibre:
Hi @yohplala Thank you for your ideas.
I have now some free time to investigate this issue.
I am currently checking we could get the assets and pairs from the CoinGecko API...

yohplala:
Hi @olibre 
One thing I have uncorrectly read in your 1st message. The quote assets, you get them with 
https://www.coingecko.com/api/documentations/v3#/simple/get_simple_supported_vs_currencies
But when you do all combinations, this remains a lot.
This is why instead of generating all possible pairs, only checking `base_asset` & `quote_asset` separately is more efficient.

olibre:
Thanks @yohplala Yes I think we will do some thing in this way...

On my side I think we can use the endpoint `supported_vs_currencies` to get the list of the base assets. This is my try using API v3:
```
curl https://api.coingecko.com/api/v3/simple/supported_vs_currencies > BaseAssets.json
jq '.[]' < BaseAssets.json | tr -d '"' | tr a-z A-Z > BaseAssets.normalized
```

yohplala:
> Thanks @yohplala Yes I think we will do some thing in this way...
> 
> On my side I think we can use the endpoint `supported_vs_currencies` to get the list of the base assets. This is my try using API v3:
> 
> ```
> curl https://api.coingecko.com/api/v3/simple/supported_vs_currencies > BaseAssets.json
> jq '.[]' < BaseAssets.json | tr -d '"' | tr a-z A-Z > BaseAssets.normalized
> ```

supported vs currencies gives you quote assets, not base assets.

olibre:
I sometimes confuse Quote and Base :facepalm:

olibre:
To clarify the situation, I will send another PR to fix the regression I have introduced with this PR. I see two solutions:
1. I simply revert the heart of `coingecko_pairs()`
2. I propose something different to better fit all airs combinations

I am currently working on the second solution. But if I do not find anything better, I will just stick on the first solution.

@yohplala has also good ideas.

olibre:
Hi @yohplala 
For your information, I have just initiated a discussion about our subject: https://github.com/bmoscon/cryptofeed/discussions/387#discussioncomment-268092

olibre:
Hi @bmoscon and @yohplala

Sorry, I thought I had introduced a bug (regression), but it was an earlier commit that changed the `coingecko_pairs()` body https://github.com/bmoscon/cryptofeed/commit/8970107

```diff
def coingecko_pairs():
-    quote_c = requests.get('https://api.coingecko.com/api/v3/coins/list').json()
-    # Normalization
-    normalized = dict({'miota': 'iota'})
-    # Base currencies are defined manually (USD + BTC + ETH).
-    # The full list from Coingecko is not used, as the generated dict of pairs would be tremendous.
-    # base_c =  requests.get('https://api.coingecko.com/api/v3/simple/supported_vs_currencies').json()
-    base_c = (('USD', 'usd'), ('BTC', 'btc'), ('ETH', 'eth'))
-    return {(f"{q['symbol']}{PAIR_SEP}{bk}".upper() if q['symbol'] not in normalized else f"{normalized[q['symbol']]}{PAIR_SEP}{bk}".upper()): f"{q['id']}_{bv}" for q in quote_c for bk, bv in base_c}
+    data = requests.get('https://api.coingecko.com/api/v3/coins/list').json()
+    return {entry['symbol'].upper(): entry['id'] for entry in data}
```

The latter `coingecko_pairs()` has two issues that I will detail in the discussion https://github.com/bmoscon/cryptofeed/discussions/387

PR #379, opened 2021-01-06
Support BITMEX authentication

olibre:
### Authenticate with API key/secret on websocket connection

Content:
- Use the `Config` feature to get the API key & secret
- Also use env. vars as a priority over Config (`Config` is the fallback)
- Add a new demo file to give an example of the different Config usages (added value)

Difficulties:
- I spend hours to figure out the missing `config=config` when creating the `Bitmex` feed
- I will send you another PR to provide feedback about Config origin to improve dev. experience

Done:
- [x] Tested
- [x] Changelog updated
- [ ] Tests run and pass
- [ ] Flake8 run and all errors/warnings resolved

### DIFF
--- a/cryptofeed/exchange/bitmex.py
+++ b/cryptofeed/exchange/bitmex.py
@@ -487,12 +487,12 @@ class Bitmex(Feed):
             elif msg['table'] == 'liquidation':
                 await self._liquidation(msg, timestamp)
             else:
-                LOG.warning("%s: Unhandled message %s", conn.uuid, msg)
+                LOG.warning("%s: Unhandled table=%r in %r", conn.uuid, msg['table'], msg)
         elif 'info' in msg:
-            LOG.info("%s - info message: %s", conn.uuid, msg)
+            LOG.info("%s: Info message from exchange: %s", conn.uuid, msg)
         elif 'subscribe' in msg:
             if not msg['success']:
-                LOG.error("%s: subscribe failed: %s", conn.uuid, msg)
+                LOG.error("%s: Subscribe failure: %s", conn.uuid, msg)
         elif 'error' in msg:
             LOG.error("%s: Error message from exchange: %s", conn.uuid, msg)
         elif 'request' in msg:
@@ -501,7 +501,7 @@ class Bitmex(Feed):
             else:
                 LOG.warning("%s: Failure %s", conn.uuid, msg['request'])
         else:
-            LOG.warning("%s: Unhandled message %s", conn.uuid, msg)
+            LOG.warning("%s: Unexpected message from exchange: %s", conn.uuid, msg)
 
     async def subscribe(self, websocket):
         self._reset()
--- a/cryptofeed/exchange/bitmex.py
+++ b/cryptofeed/exchange/bitmex.py
@@ -473,17 +473,17 @@ class Bitmex(Feed):
 
         msg = json.loads(msg, parse_float=Decimal)
         if 'info' in msg:
-            LOG.info("%s - info message: %s", self.id, msg)
+            LOG.info("%s - info message: %s", conn.uuid, msg)
         elif 'subscribe' in msg:
             if not msg['success']:
-                LOG.error("%s: subscribe failed: %s", self.id, msg)
+                LOG.error("%s: subscribe failed: %s", conn.uuid, msg)
         elif 'error' in msg:
-            LOG.error("%s: Error message from exchange: %s", self.id, msg)
+            LOG.error("%s: Error message from exchange: %s", conn.uuid, msg)
         elif 'request' in msg:
             if msg['success']:
-                LOG.info("%s: Success %s", self.id, msg['request'].get('op'))
+                LOG.info("%s: Success %s", conn.uuid, msg['request'].get('op'))
             else:
-                LOG.warning("%s: Failure %s", self.id, msg['request'])
+                LOG.warning("%s: Failure %s", conn.uuid, msg['request'])
         elif 'table' in msg:
             if msg['table'] == 'trade':
                 await self._trade(msg, timestamp)
@@ -498,9 +498,9 @@ class Bitmex(Feed):
             elif msg['table'] == 'liquidation':
                 await self._liquidation(msg, timestamp)
             else:
-                LOG.warning("%s: Unhandled message %s", self.id, msg)
+                LOG.warning("%s: Unhandled message %s", conn.uuid, msg)
         else:
-            LOG.warning("%s: Unhandled message %s", self.id, msg)
+            LOG.warning("%s: Unhandled message %s", conn.uuid, msg)
 
     async def subscribe(self, websocket):
         self._reset()
--- a/cryptofeed/exchange/bitmex.py
+++ b/cryptofeed/exchange/bitmex.py
@@ -4,10 +4,15 @@ Copyright (C) 2017-2021  Bryant Moscon - bmoscon@gmail.com
 Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
-import logging
 from collections import defaultdict
 from datetime import datetime as dt
 from decimal import Decimal
+import hashlib
+import hmac
+import logging
+import os
+import time
+import urllib.parse
 
 import requests
 from sortedcontainers import SortedDict as sd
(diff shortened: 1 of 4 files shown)


cryptofeed/exchange/bitmex.py
```diff
@@ -502,3 +508,36 @@ async def subscribe(self, websocket):
...
             await websocket.send(json.dumps({"op": "subscribe",
                                              "args": chans[i:i + 10]}))
+
+    @staticmethod
+    async def _authenticate(conn):
+        """Send API Key with signed message.
+
+        https://github.com/BitMEX/sample-market-maker/blob/master/test/websocket-apikey-auth-test.py"""
+        BITMEX_API_KEY = os.environ.get("BITMEX_API_KEY")
+        BITMEX_API_SECRET = os.environ.get("BITMEX_API_SECRET")
```
bmoscon, line 518: there is a config file that should be used for this now

cryptofeed/exchange/bitmex.py
```diff
@@ -502,3 +508,36 @@ async def subscribe(self, websocket):
...
+
+        if not BITMEX_API_KEY or not BITMEX_API_SECRET:
+            LOG.warning("BITMEX: no authentication - missing environment variables: BITMEX_API_KEY and BITMEX_API_SECRET")
```
bmoscon, line 521: inside the exchange files when referencing the exchange name please use self.id. Currently they are all just uppercase strings, but should the format change i dont want to have to hunt down aberrations like this

bmoscon (changes requested):

cryptofeed/exchange/bitmex.py
olibre: Done :+1:

cryptofeed/exchange/bitmex.py
olibre: I know the config file to list the channels and pairs, but not sure what you mean by using the config file to store the API_KEY and API_SECRET. :thinking:  I do not use (yet) the config file.

I have added two arguments: `key_api` and `key_secret`.
And removed the arguments `pairs=None, channels=None, callbacks=None` because I do not see why we require them: we already pass `kwargs` to `super.__init__()` isn't it?

If my change breaks something, please implement what you want yourself, and I will learn what you was expecting.

bmoscon: I think it would be better if feed's were initialized with config objects directly from the feedhandler. None of this has been done yet, I only added config file support in feedhandler. I think ideally the feedhandler can either pass the entire config object through to each feed, or it can use the feed id to identify sections in the config to pass through. I'm open to suggestions, but i def do not want to support a million ways to do it. function args, or config or env vars is just too much. The config file can be specified via an env var, thats probably sufficient customization

bmoscon: the config file is not the same as the channels/pairs - the config file currently has the logging settings in it only, but in the future it will have exchange specific options (like key, secret)

bmoscon: what do you think about config file and env var, or config file and kwargs? I think supporting 2 of the 3 is fine, but all 3 seems like a hassle and overkill

olibre: On my opinion, there are four kinds of settings origin:
1. Users using Cryptofeed as a third party, for example, with other tools as Cryptostore. I think these users are happy with a configuration file = they do not want to code the configuration, just parameterize it.
2. Users loving shell scripting and CLI. They are used with command line arguments. However, we can skip this case as Cryptofeed is not a standalone application (not yet?).
3. Users using Cryptofeed as a library, like in examples/demo_xxx.py. I think some of these users want to just hard code their own requirements in their Python source code.
4. Users putting Cryptofeed within a container. These users are used to configure software with environment vars.
5. There is a fifth case: settings from remote database. The users set the configuration with a (web) front-end. Let's ignore that fifth case.

I am in the last case: I prefer to edit the software configuration using environment variables in my container configuration file. 

To answer to your question, I depends on the user experience you want and on the simplicity/complexity you want in Cryptofeed source code.

I have already implemented software having different settings origins:
1. CLI arguments (priority)
2. env. var.
3. config fie
4. least priority: default values. 

Printing the origin of the settings is important for the user experience, because the user may spend hours before understanding that that command flag or env. var. is overridden the settings they want.

I think we should search for a Python lib that already implements that requirement: upload a configuration file, overwrite the values from environment variables and then checking the `kwargs`. Implementing this for Cryptofeed would take a couple of days.

What is your preference?
1. searching Python libraries / comparing / integrating it with Cryptofeed (a couple of days)
2. implementing a tiny/simple class that does it for our own needs (a couple of days)
3. doing both and keeping the best

olibre: How easy do you think we can introduce `os.environ.get()` when doing the following?
```py
with open(config_file) as fp:
    self.config = AttrDict(yaml.safe_load(fp))
```

The value of `self.config.log.level` should be overwritten by `os.environ.get(f'{prefix}_LOG_LEVEL')` if that last env. var. is found.

The `prefix` is set by the owner of the configuration. For example the `FeedHandler` may set `prefix='CRYPTOFEED_FEEDHANDLER'`.

The function `add_feed()` can also automatically overwrite the configuration with the passed `pairs`, `channels` and `kwargs`.

### Example

For example, the config says something like:
```yaml
bitmex:
  api_key: 123-file-456789
```

The env. var. `CRYPTOFEED_FEEDHANDLER_BITMEX_API_KEY` says `ABC-env-DEFG` and overwrite the value from the config file `123-file-456789`.

Therefore, at the end of the `FeedHandler.__init__()` we will get `self.config.bitmex.key_value='123456789'`.

But the `Bitmex` class is created with `Bitmex(api_key='foobar-kwarg')` overwriting the previous value `ABC-env-DEFG`. Thus, in the `Bitmex` source code, we will have `self.config.api_key='foobar-kwarg'` and the same value in the rest of the `FeedHandler` source code.

In lieu of parsing `self.config` and then `self.channels`, we only parse `self.config`.

I think we can implement something simple & automagical and make all users happy.

bmoscon: @olibre take a look at my commit, specifically this doc: https://github.com/bmoscon/cryptofeed/blob/master/docs/config.md

I what it suggests is pretty reasonable, let me know

bmoscon: this is the commit in question. Since i renamed a kwarg, its pretty large:

https://github.com/bmoscon/cryptofeed/commit/d81993957b1847879793c89957e32c8e35620d35

olibre: Great commit. Thank you for clarifying the config file. Let me think how to magically add support for env. vars... :thinking:

bmoscon: @olibre sure and the commit is just a suggestion - a way to get something in that we can use as a basis for any other changes. I'm open to suggestions, but in my opinion, env vars are not really feasible, you'll need 2-4 per exchange potentially, plus some high level ones, so we're looking at over 100 env vars which doesnt seem like a good idea. I use containers in cloud settings daily and we use config files and env vars typically. I'm open to improvements and suggestions! Just wanted to set a foundation, even if its un-ideal - it can always be changed :D

olibre: I thought of env. vars only to override a few settings, like for a few API keys/ids/secrets. I think this is what your company does with containers in cloud, isn't it?

bmoscon: well my experience with env vars is that a lot of managed service offerings from azure, aws, gcp, etc require them for specific settings, so I guess the answer to your question is yes. I guarantee if one exchange lets you override the keys with env vars, someone is going to want that for some other exchange, and at that point you might as well have them for everything. I think right now then only place env vars are used is to override the location of a config file, but i'm open to ideas

bmoscon: @olibre my proposal for now would be for me to accept this PR and integrate it in with the way I've set up the config files while you think about changes to the overall config architecture. How does that sound?

olibre: OK I understand your intention. I am OK to adapt this current PR in order to use your proposal. I will take care of that later. :+1:

bmoscon: @olibre let me know if you want me to handle the tweaks post merge

olibre: Hi Bryant
I have just finished the PR about CoinGecko duplicated/collided symbols.
I think I will manage this PR in one hour, but I need some sleep before...
Take care, cheers :wink:

olibre: New code is using `Config` feature => Bryant, you can review again this PR.

bmoscon (approved):

PR #383, opened 2021-01-06
Avoid using SIGHUP on ImportError

olibre:
Add variable SIGNALS to avoid using SIGHUP in source code when `from signal import SIGHUP` fails.


bmoscon (approved):
good catch :)

PR #390, opened 2021-01-13
Add guard against non-supported asyncio add_signal_handler() on windows platforms

davidtwomey:

Windows does not support asyncio loop.add_signal_handler() (See: https://docs.python.org/3/library/asyncio-platforms.html)

Get a NotImplementedError when running from windows.

This PR just adds a simple check if we are on a windows platform to avoid adding signal handles.

- [x] - Tested
- [x] - Changelog updated
- [ ] - Tests run and pass

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,7 +16,8 @@
   * Breaking change: Feed object `config` renamed `subscription`
   * Feature: Configuration passed from feedhandler to exchanges
   * Breaking change: most use of `pair` and `pairs` changed to `symbol` and `symbols` to be more consistent with actual usage. pairs.py renamed to symbols.py
-
+  * Bugfix: Add guard against non-supported asyncio add_signal_handler() on windows platforms
+    
 ### 1.6.2 (2020-12-25)
   * Feature: Support for Coingecko aggregated data per coin, to be used with a new data channel 'profile'
   * Feature: Support for Whale Alert on-chain transaction data per coin, to be used with a new data channel 'transactions'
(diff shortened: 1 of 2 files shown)


olibre:
Hi @davidtwomey 
I cannot test Cryptofeed using Windows but I have read this SO answer:
https://stackoverflow.com/a/54886771

Maybe a better workaround is to use the classic function `signal.signal()` on Windows.

You may test the following code:

```py
for signal in SIGNALS:
    signal.signal(signal, handle_stop_signals)
```

in lieu of the current one:

```py
for signal in SIGNALS:
    loop.add_signal_handler(signal, handle_stop_signals)
```

Please give us some feedback. :wink:

davidtwomey:
Hi @olibre,

Yes that could work! 
Are you proposing to a) switch completely to `signal.signal` or b) only if on a windows (and therefore keep the `loop.add_signal_handler` generally)?

bmoscon:
@davidtwomey in the conditional you added you can probably just substitute 

```
for signal in SIGNALS:
    signal.signal(signal, handle_stop_signals)
```
for the return statement you have

olibre:
Hi @davidtwomey 
I was thinking b) keep the `loop.add_signal_handler()` generally but use `signal.signal()` only if on windows :+1:

davidtwomey:
Ok `signal.signal` is now implemented for windows. 

Note `handle_stop_signals` has to now accept *args or you get the following (on windows):

```
[ERROR] Encountered error handle_stop_signals() takes 0 positional arguments but 2 were given...closing
```

bmoscon (approved):

bmoscon:
thanks for the PR!

PR #392, opened 2021-01-14
Provide Config origin feedback to developers

olibre:
### Improve developer experience by printing where the Config is coming from

- [ ] Tested
- [x] Changelog updated
- [ ] Tests run and pass
- [ ] Flake8 run and all errors/warnings resolved

### DIFF
--- a/cryptofeed/config.py
+++ b/cryptofeed/config.py
@@ -5,7 +5,6 @@ Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
 import os
-import sys
 
 import yaml
 
@@ -36,6 +35,8 @@ class AttrDict(dict):
 
 class Config:
     def __init__(self, config=None):
+        self.config = AttrDict(_default_config)
+        self.log_msg = ""
         if isinstance(config, str):
             config_file = os.path.join(os.path.dirname(os.path.abspath(__file__)), "config.yaml")
             if config and os.path.exists(config):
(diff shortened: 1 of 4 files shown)


cryptofeed/config.py
```diff
@@ -25,9 +26,11 @@ def __setitem__(self, key, value):
         super().__setitem__(key, value)
 
     def __getattr__(self, item):
+        # print(f'Config: found {item!r}', file=sys.stderr, flush=True)
```
bmoscon, line 29: did you mean to leave these commented out?

bmoscon: i'd think we should just remove them

olibre: Done :+1:

bmoscon (commented):
i was just thinking, the feedhandler doesnt init the logger until after it processes the log file - does this work (by that I mean does it log to the file correctly, based on the user specified log levels)?

olibre:
> i was just thinking, the feedhandler doesnt init the logger until after it processes the log file - does this work (by that I mean does it log to the file correctly, based on the user specified log levels)?

I am not sure what you wonder. I guess you are speaking about the `LOG.info()` in the `Feed.__init__()`. I was hesitating between `LOG.info()` and `print()`. For that matter, my first implementation used `print()`.

Let's compare three use cases:

1. When `add_feed()` is used as the following, the `FeedHandler` creates the logger before creating the `Feed` sub-class, and the `Feed.__init__()` logs correctly.

	  ```py
	  f = FeedHandler()
	  f.add_feed('BITMEX', symbols=['XBTUSD'], channels=[L2_BOOK], callbacks={L2_BOOK: my_function})
	  ```

2. In this second case the `Feed` sub-class created before the `add_feed()` is called, but the `FeedHandler` has already created the logger, and then the `Feed.__init__()` logs correctly.

	  ```py
	  f = FeedHandler()
	  f.add_feed(Bitmex(symbols=['XBTUSD'], channels=[L2_BOOK], callbacks={L2_BOOK: my_function}))
	  ```

3. When the `Feed` sub class is created before the `FeedHandler` as in the following snippet, we may have an issue about the `LOG.ingo()` in the `Feed.__init__()`. 

	  ```py
      b = Bitmex(symbols=['XBTUSD'], channels=[L2_BOOK], callbacks={L2_BOOK: my_function})
	  f = FeedHandler()
	  f.add_feed(b)
	  ```

In that last case, I think this log is not printed, but not sure because I have not tested it. In fact, I do not care for these reasons:
1. This is a corner case, not sure if many people create the `Feed` sub-classes before the `FeedHandler`.
2. This log line is not important, and missing it is not a big issue.
3. Better to use `LOG.info()` in lieu of `print()`

You know why I have preferred to use `Log.info()` in lieu of `print()` in `Feed.__init__()`.
Is it OK for you?

bmoscon:
my thought was more about these lines in the feedhandler

```python

       self.config = Config(config=config)

        get_logger('feedhandler', self.config.log.filename, self.config.log.level)

```

the logger is not configured until after the config is created, so that means any log lines in config are just going to the default logger (which I think is stdout?)

bmoscon:
I see, thats why you have prints there. I'm very strongly anti-print statement. I dont like things emitting random output to stdout; especially not a library.

bmoscon:
could we move that logging out of config and into feedhandler (and log it to a log file at the debug level?), or, log to a separate file? or something else? just thinking out loud

bmoscon (approved):

PR #393, opened 2021-01-14
Get API KEY ID from env. var or from Config

olibre:
### Give flexibility to use Config or env. var. to pass secrets

The environment variable format is: `CF_{exchange-name}_KEY_ID`

The env. var. has a higher priority.

This feature will be explained in documentation if this PR would be merged.

This is a tiny one-line change consistent with the other PR about Bitmex authentication.
No feedback is provided to trace the origin of the used API KEY ID (env. var. or Config).

To be more consistent across source code, I can create another PR to move all related Config/env.var. values is a generic function in Config, with the same behavior. But this enhancement would be a future PR. First let's implement it in a KISS fashion.

- [ ] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass

### DIFF
--- a/cryptofeed/feed.py
+++ b/cryptofeed/feed.py
@@ -5,6 +5,7 @@ Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
 from collections import defaultdict
+import os
 from typing import Tuple, Callable, Union, List
 
 from cryptofeed.callback import Callback
@@ -57,7 +58,7 @@ class Feed:
         self.origin = origin
         self.checksum_validation = checksum_validation
         self.ws_defaults = {'ping_interval': 10, 'ping_timeout': None, 'max_size': 2**23, 'max_queue': None, 'origin': self.origin}
-        key_id = self.config[self.id.lower()].key_id
+        key_id = os.environ.get(f'CF_{self.id}_KEY_ID') or self.config[self.id.lower()].key_id
         load_exchange_symbol_mapping(self.id, key_id=key_id)
 
         if subscription is not None and (symbols is not None or channels is not None):


bmoscon (approved):

bmoscon:
seems reasonable. the key id is only a part of what is needed though, right?

bmoscon (approved):

olibre:
> seems reasonable. the key id is only a part of what is needed though, right?

Yes this PR is only about API KEY ID.

Nevertheless, we can imagine another PR that proposes to generalize this env-var-or-config feature by adding a function in `config.py`. If you want I can think about it :thinking: and send you another PR later...

nirvana-msu:
@bmoscon I can see this feature (providing API key/secret via env vars) was removed in [subsequent refactoring](https://github.com/bmoscon/cryptofeed/commit/b02122e070c14c8507504f6f80f24ce653b5d7e2#diff-b93bc2e8c87b929ee0b7ec36f92224c97202e779565241eeb247ddda8e5801d8L179). Can't find any mention of this feature being removed in changelog, or the reason for removal. Also, Bitmex demo file [still references](https://github.com/bmoscon/cryptofeed/blob/cb4c2f7ceebe942dd5d5f2fb9e7090f220e14754/examples/demo_bitmex_config.py#L34-L35) these env vars which doesn't work anymore.

Was there a particular reason for getting this feature removed, or is it just a collateral of refactoring? Can we get it back, please?

bmoscon:
@nirvana-msu - thanks for the heads up - I've removed the superfluous example code. If you want to use ENV vars, nothing is stopping you from doing that. When you create a config dict you can embed your environment variables in there (eg load them out with os.environ, etc)

PR #405, opened 2021-01-30
SSL option for arctic callbacks

ph4z:
When you create an arctic callback instance it is actually not possible to enable SSL option.
This PR add an ssl boolean argument to the arctic callback.

Using kwargs seems not possible as it is used further in the code. 
Additional work could be needed to support other options.

- [x] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass

### DIFF
--- a/cryptofeed/backends/arctic.py
+++ b/cryptofeed/backends/arctic.py
@@ -14,7 +14,7 @@ from cryptofeed.defines import FUNDING, OPEN_INTEREST, TICKER, TRADES, LIQUIDATI
 
 
 class ArcticCallback:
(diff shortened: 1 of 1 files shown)


bmoscon (changes requested):
isnt the pymongo kwarg tls? https://pymongo.readthedocs.io/en/stable/examples/tls.html

ph4z:
I just tried the tls option like this: `LiquidationsArctic('cryptofeed-liquidations', tls=True)`
It doesn't work.

bmoscon (approved):

PR #428, opened 2021-03-04
Set default l2 depth for Kraken

dirtyValera:
This is to fix https://github.com/bmoscon/cryptostore/issues/163

Tested on mac


bmoscon:
I feel like the correct way to handle this would be on the cryptostore side. If max depth is not set in the config, it can be omitted from the kwargs

bmoscon:
going to close this - let me know if you have any questions about the comment above @dirtyValera

Issue #431, opened 2021-03-07
Enhance Speed of Huobi Spot L2_BOOK


mutedstack:
**Is your feature request related to a problem? Please describe.**
Now cryptofeed is using [market-depth](https://huobiapi.github.io/docs/spot/v1/en/#market-depth) which is snapshot at 1 second interval to fetch L2_BOOK.

**Describe the solution you'd like**
Use [market-by-price-incremental-update](https://huobiapi.github.io/docs/spot/v1/en/#market-by-price-incremental-update) whose latency is below 100ms to fetch delta book and update local cache according to it

**Describe alternatives you've considered**
Use [market-by-price-refresh-update](https://huobiapi.github.io/docs/spot/v1/en/#market-by-price-refresh-update) which will push full order book below 100ms but it's slower than delta book above.

**Additional context**
Market-by-price-incremental-update in official Python SDK: https://github.com/HuobiRDCenter/huobi_Python/blob/master/huobi/service/market/sub_mbp_increase.py#L8

zhenpingfeng:
Can you help me test the incremental version I implemented?

PR #432, opened 2021-03-09
Rewrite Huobi to adapt market-by-price-incremental-update.

zhenpingfeng:
[Rewrite Huobi to adapt market-by-price-incremental-update and impelemet Huobi's Ticker channel.](https://github.com/bmoscon/cryptofeed/issues/431#issue-823895060)
- [ ] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass

Shouldn't be merge because not fully test yet, need some review.


zhenpingfeng:
When the network delay is high, it is difficult to match prevSeqNum with new seqNum. In high delay of network, some data will always be missed. I will turn off this pr first, and then submit the pr after I do it right.

PR #452, opened 2021-03-25
support for authenticated channel on Okex (and by default OKCoin)

twmeggs:

- [ ] - Adds websocket authentication for Okex and, by default, OKCoin
- [ ] - Handles separately required login and subscription
- [ ] - Only implemented for `spot/order` channel and hardcoded for symbol `BTC-USDT` at present (see: `auth_subscribe` function in cryptofeed/exchange/okcoin)
- [ ] - see the working example in `examples/demo_okex_authenticated.py`

### DIFF
--- a/cryptofeed/exchange/okcoin.py
+++ b/cryptofeed/exchange/okcoin.py
@@ -4,19 +4,25 @@ Copyright (C) 2017-2021  Bryant Moscon - bmoscon@gmail.com
 Please see the LICENSE file for the terms and conditions
 associated with this software.
 '''
+import asyncio
+import time
+
 from decimal import Decimal
 from itertools import islice
 import logging
+import websockets
 import zlib
+from typing import List, Tuple, Callable
 
 from sortedcontainers import SortedDict as sd
 from yapic import json
 
+from cryptofeed.auth.okcoin import generate_token
 from cryptofeed.connection import AsyncConnection
-from cryptofeed.defines import ASK, BID, BUY, FUNDING, L2_BOOK, OKCOIN, OPEN_INTEREST, SELL, TICKER, TRADES, LIQUIDATIONS
+from cryptofeed.defines import ASK, BID, BUY, FUNDING, L2_BOOK, OKCOIN, OPEN_INTEREST, SELL, TICKER, TRADES, LIQUIDATIONS, ORDER_INFO
 from cryptofeed.exceptions import BadChecksum
 from cryptofeed.feed import Feed
-from cryptofeed.standards import symbol_exchange_to_std, timestamp_normalize
+from cryptofeed.standards import symbol_exchange_to_std, timestamp_normalize, is_authenticated_channel
 from cryptofeed.symbols import get_symbol_separator
 from cryptofeed.util import split
 
@@ -70,6 +76,7 @@ class OKCoin(Feed):
         for chunk in split.list_by_max_items(symbol_channels, 33):
             LOG.info("%s: Subscribe to %s args from %r to %r", self.id, len(chunk), chunk[0], chunk[-1])
             request = {"op": "subscribe", "args": chunk}
+            print('request is:', request)
             await conn.send(json.dumps(request))
 
     @staticmethod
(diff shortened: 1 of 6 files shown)


cryptofeed/exchange/okcoin.py
```diff
@@ -181,6 +189,30 @@ async def _book(self, msg: dict, timestamp: float):
...
+                            order_id=msg['data'][0]['order_id'],
+                            side=BUY if msg['data'][0]['side'].lower() == 'buy' else SELL,
+                            order_type=msg['data'][0]['order_type'],
+                            timestamp=msg['data'][0]['timestamp'].timestamp(),
+                            receipt_timestamp=timestamp,
+                            **data
+                            )
+        
+    async def _login(self, msg: dict, timestamp: float):
+        print(f"Login result: {msg}")
```
bmoscon, line 215: No print statements. If you want, you can log this as info or debug (there are logging examples throughout)

cryptofeed/exchange/okcoin.py
```diff
@@ -204,7 +242,30 @@ async def message_handler(self, msg: str, conn, timestamp: float):
...
+        return ret
+    
+    async def auth_subscribe(self, conn: AsyncConnection, options=None):
+        self.__reset()
+        timestamp, sign = generate_token(self.key_id, self.key_secret)            
+        login_param = {"op": "login", "args": [self.key_id, self.config.okex.key_passphrase, timestamp, sign.decode("utf-8")]}
+        login_str = json.dumps(login_param)
+        await conn.send(login_str)
+        await asyncio.sleep(5)
+        sub_param = {"op": "subscribe", "args": ["spot/order:BTC-USDT"]}
```
bmoscon, line 269: this is hard coded to only BTC, right, not what the user has configured?

bmoscon (changes requested):

cryptofeed/exchange/okcoin.py
twmeggs: Removed in second commit

cryptofeed/exchange/okcoin.py
twmeggs: updated and improved in second commit - thanks for the comments and suggestions

cryptofeed/exchange/okcoin.py
```diff
@@ -70,6 +77,7 @@ async def subscribe(self, conn: AsyncConnection):
         for chunk in split.list_by_max_items(symbol_channels, 33):
             LOG.info("%s: Subscribe to %s args from %r to %r", self.id, len(chunk), chunk[0], chunk[-1])
             request = {"op": "subscribe", "args": chunk}
+            print('request is:', request)
```
bmoscon, line 80: this needs to be removed

cryptofeed/exchange/okcoin.py
```diff
@@ -204,7 +240,34 @@ async def message_handler(self, msg: str, conn, timestamp: float):
...
                 LOG.warning("%s: Unhandled message %s", self.id, msg)
         else:
             LOG.warning("%s: Unhandled message %s", self.id, msg)
+    
+    def connect(self) -> List[Tuple[AsyncConnection, Callable[[None], None], Callable[[str, float], None]]]:
+        ret = []
+        for channel in self.subscription or self.channels:
+            if is_authenticated_channel(channel):
+                if channel in self.subscription:
+                    ret.append((AsyncConnection(self.address, self.id, **self.ws_defaults), partial(self.user_order_subscribe, symbol=self.subscription.get(channel).pop()), self.message_handler))
```
bmoscon, line 255: it looks like authenticated subscribe only takes 1 symbol. If you pop off the contents of the sub dict you could have a list of symbols. If you pop it off, on reconnects, it will be lost. I think you can make it something like this:

```python
if is_authenticated_channel(channel):
    syms = self.symbols or self.subscription[channel]
   for s in syms:
       ret.append((AsyncConnection(self.address, self.id, **self.ws_defaults), partial(self.user_order_subscribe, symbol=s), self.message_handler))
```

Thats obviously untested, so if you use that test it first

bmoscon (changes requested):
see new comments

cryptofeed/exchange/okcoin.py
twmeggs: incorporated - thanks

cryptofeed/exchange/okcoin.py
twmeggs: done

bmoscon (approved):

PR #473, opened 2021-04-15
Uniform arguments

OnlyC:
- Add feed & symbol to data in BackendBookCallback & BackendBookDeltaCallback
- Changed id in BackendTradeCallback's data to order_id  to match with order_type & order_id of BackendLiquidationsCallback.

### DIFF
--- a/cryptofeed/backends/backend.py
+++ b/cryptofeed/backends/backend.py
@@ -48,14 +48,14 @@ class BackendQueue:
 
 class BackendBookCallback:
     async def __call__(self, *, feed: str, symbol: str, book: dict, timestamp: float, receipt_timestamp: float):
-        data = {'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': False, BID: {}, ASK: {}}
+        data = {'feed': feed, 'symbol': symbol, 'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': False, BID: {}, ASK: {}}
         book_convert(book, data, convert=self.numeric_type)
         await self.write(feed, symbol, timestamp, receipt_timestamp, data)
 
 
 class BackendBookDeltaCallback:
     async def __call__(self, *, feed: str, symbol: str, delta: dict, timestamp: float, receipt_timestamp: float):
-        data = {'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': True, BID: {}, ASK: {}}
+        data = {'feed': feed, 'symbol': symbol, 'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': True, BID: {}, ASK: {}}
         book_delta_convert(delta, data, convert=self.numeric_type)
         await self.write(feed, symbol, timestamp, receipt_timestamp, data)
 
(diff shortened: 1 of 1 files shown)


cryptofeed/backends/backend.py
```diff
@@ -48,22 +48,22 @@ async def read_many_queue(self, count: int):
...
+        data = {'feed': feed, 'symbol': symbol, 'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': True, BID: {}, ASK: {}}
         book_delta_convert(delta, data, convert=self.numeric_type)
         await self.write(feed, symbol, timestamp, receipt_timestamp, data)
 
 
 class BackendTradeCallback:
     async def __call__(self, *, feed: str, symbol: str, side: str, amount: Decimal, price: Decimal, order_id: str = None, timestamp: float, receipt_timestamp: float, order_type: str = None):
         data = {'feed': feed, 'symbol': symbol, 'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp,
-                'side': side, 'amount': self.numeric_type(amount), 'price': self.numeric_type(price), 'order_type': order_type, 'id': order_id}
+                'side': side, 'amount': self.numeric_type(amount), 'price': self.numeric_type(price), 'order_type': order_type, 'order_id': order_id}
```
bmoscon, line 66: you'll need to change a lot more code if you change id to order id. That specific key is used elsewhere

bmoscon (changes requested):

cryptofeed/backends/backend.py
```diff
@@ -48,22 +48,22 @@ async def read_many_queue(self, count: int):
 
 class BackendBookCallback:
     async def __call__(self, *, feed: str, symbol: str, book: dict, timestamp: float, receipt_timestamp: float):
-        data = {'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': False, BID: {}, ASK: {}}
+        data = {'feed': feed, 'symbol': symbol, 'timestamp': timestamp, 'receipt_timestamp': receipt_timestamp, 'delta': False, BID: {}, ASK: {}}
```
bmoscon, line 51: that data is already present in the args passed to write. why do you think we need it here? Most the backends that use this will either ignore that or dump and store the json as-is, which would lead to a lot of duplicated entries (and wasted storage) since book data is so voluminous

cryptofeed/backends/backend.py
OnlyC: This is just to make the parameter names are match in all callbacks. If not we can still use it.

cryptofeed/backends/backend.py
OnlyC: Because I the data doesn't include the symbol and I need that in callback function. Right now, I have to use 
`key.rsplit('-', 1)[-1]`
to get the symbol.

Other callbacks have feed & symbol in data. If we choose to get rid of duplicated data, shouldn't we remove them from all the callbacks?

cryptofeed/backends/backend.py
bmoscon: I'm fine with it changing to order_id, but if you look in the other backend code some of them expect a field called "id" so you'll need to update all of those

bmoscon:
hasnt been any activity on this in a while. if you want to continue with it at some point, please re-open/recreate

PR #474, opened 2021-04-17
Add argument to specify the Postgres database port

rvdmaazen:
Added an `__init__` argument to `backend.PostgresCallback` that can be used to specify the port on which to connect to the Postgres database in the case that the database port differs from the standard port `5432`.

### DIFF
--- a/cryptofeed/backends/postgres.py
+++ b/cryptofeed/backends/postgres.py
@@ -18,7 +18,7 @@ from cryptofeed.defines import CANDLES, FUNDING, OPEN_INTEREST, TICKER, TRADES, 
 
 
 class PostgresCallback(BackendQueue):
-    def __init__(self, host='127.0.0.1', user=None, pw=None, db=None, table=None, numeric_type=float, max_batch=100, **kwargs):
+    def __init__(self, host='127.0.0.1', user=None, pw=None, db=None, port=None, table=None, numeric_type=float, max_batch=100, **kwargs):
         """
         host: str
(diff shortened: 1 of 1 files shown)


cryptofeed/backends/postgres.py
```diff
@@ -40,11 +40,12 @@ def __init__(self, host='127.0.0.1', user=None, pw=None, db=None, table=None, nu
         self.db = db
         self.pw = pw
         self.host = host
+        self.port = port
         self.max_batch = max_batch
 
     async def _connect(self):
         if self.conn is None:
-            self.conn = await asyncpg.connect(user=self.user, password=self.pw, database=self.db, host=self.host)
+            self.conn = await asyncpg.connect(user=self.user, password=self.pw, database=self.db, host=self.host, port=self.port)
```
bmoscon, line 48: will this work if its none? Shouldnt the default be the default Postgres port, 5432?

bmoscon (changes requested):

cryptofeed/backends/postgres.py
rvdmaazen: It will work if `None` will be passed to `asyncpg.connect` since this is also the default value, see also [their documentation](https://magicstack.github.io/asyncpg/current/api/index.html#connection). It will automatically fall back to port `5432` if no port is provided using the `port` argument or the `PGPORT` environment variable.

bmoscon (approved):

PR #475, opened 2021-04-22
Update huobi.py

wufantastic:
fixed amount bug


bmoscon (approved):

bmoscon:
good catch. We def want base vol, not quote vol

PR #483, opened 2021-05-02
fix pip & Kraken trade history. added ledger info

cupidsdead:
fixes pipfile dependency break
fixes get_trades_history

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -26,7 +26,7 @@ cchardet = "*"              # cchardet is a faster replacement for chardet
 aiofile = ">=2.0.0"         # used by util/async_file.py
 aiokafka = "*"              # pip install cryptofeed[kafka]
 hiredis = "*"               # pip install cryptofeed[redis]
-aioredis = ">=2.0.0a1"      # pip install cryptofeed[redis]
+aioredis = "*"      # pip install cryptofeed[redis]
 arctic = "*"                # pip install cryptofeed[arctic]
 asyncpg = "*"               # pip install cryptofeed[postgres]
 motor = "*"                 # pip install cryptofeed[mongo]
--- a/Pipfile
+++ b/Pipfile
@@ -26,7 +26,7 @@ cchardet = "*"              # cchardet is a faster replacement for chardet
 aiofile = ">=2.0.0"         # used by util/async_file.py
 aiokafka = "*"              # pip install cryptofeed[kafka]
 hiredis = "*"               # pip install cryptofeed[redis]
-aioredis = "*"      # pip install cryptofeed[redis]
+aioredis = ">=2.0.0a1"      # pip install cryptofeed[redis]
 arctic = "*"                # pip install cryptofeed[arctic]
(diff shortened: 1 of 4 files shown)


Pipfile
```diff
@@ -26,7 +26,7 @@ cchardet = "*"              # cchardet is a faster replacement for chardet
 aiofile = ">=2.0.0"         # used by util/async_file.py
 aiokafka = "*"              # pip install cryptofeed[kafka]
 hiredis = "*"               # pip install cryptofeed[redis]
-aioredis = ">=2.0.0a1"      # pip install cryptofeed[redis]
+aioredis = "*"      # pip install cryptofeed[redis]
```
bmoscon, line 29: why did you change this? it will only work with that specific version of aioredis (the pre-release alpha). Older versions will not work.

bmoscon (changes requested):

Pipfile
cupidsdead: I see. On that version it was unable to resolve dependencies and create a lockfile via Pipfile.

cryptofeed/rest/kraken.py
```diff
@@ -160,7 +160,8 @@ def trades(self, symbol: str, start=None, end=None, retry=None, retry_wait=10):
             if not end:
                 end = pd.Timestamp.utcnow()
             for data in self._historical_trades(symbol, start, end, retry, retry_wait):
-                yield list(map(lambda x: self._trade_normalization(x, symbol), data['result'][next(iter(data['result']))]))
+                yield list(
+                    map(lambda x: self._trade_normalization(x, symbol), data['result'][next(iter(data['result']))]))
```
bmoscon, line 164: looks like just a formatting change, please undo

cryptofeed/rest/kraken.py
```diff
@@ -317,3 +281,91 @@ def cancel_order(self, order_id: str):
...
+            XETH => ETH
+            ZGBP => GBP
+        """
+        cleansym = sym
+        try:
+            if len(sym) == 8:
+                cleansym = sym[1:4] + sym[5:]
+            elif len(sym) == 4:
+                cleansym = sym[1:]
+        except Exception as ex:
```
bmoscon, line 368: at a minimum log this, i can't think of any case where an except pass is a good idea, except on perhaps places where you might be trying to import various modules and might expect issues

bmoscon (changes requested):
looks pretty good, just a few minor changes :)

cryptofeed/rest/kraken.py
cupidsdead: OK this is now being logged and expanded the logic so we catch pairs like ETHUSDT

cupidsdead:
> looks pretty good, just a few minor changes :)

Thanks for checking :) have made your requested changes. are we able to merge now? Cheers.

cryptofeed/rest/kraken.py
cupidsdead: this is now resolved

bmoscon (approved):

PR #496, opened 2021-05-22
Added support for Bitmex Testnet

Pcsl:
Added the testnet URL:  'wss://testnet.bitmex.com/realtime'

Probably it won't work as more code needs to be changed. It's my first PR ever so if code is not god enough or if just doesn't work please close it (or someone else can try to fix it).

### DIFF
--- a/cryptofeed/exchange/bitmex.py
+++ b/cryptofeed/exchange/bitmex.py
@@ -56,7 +56,7 @@ class Bitmex(Feed):
 
     def __init__(self, **kwargs):
         auth_api = 'wss://www.bitmex.com/realtime' if not sandbox else 'wss://testnet.bitmex.com/realtime'
-        super().init(auth_api, **kwargs)
+        super().__init__(auth_api, **kwargs)
         self._reset()
 
     def _reset(self):
--- a/cryptofeed/exchange/bitmex.py
+++ b/cryptofeed/exchange/bitmex.py
@@ -55,7 +55,8 @@ class Bitmex(Feed):
         return ret, info
 
     def __init__(self, **kwargs):
-        super().__init__('wss://www.bitmex.com/realtime', **kwargs)
+        auth_api = 'wss://www.bitmex.com/realtime' if not sandbox else 'wss://testnet.bitmex.com/realtime'
+        super().__init__(f'{auth_api}/realtime', **kwargs)
         self._reset()
 
     def _reset(self):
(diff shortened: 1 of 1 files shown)


cryptofeed/exchange/bitmex.py
```diff
@@ -55,7 +55,8 @@ def _parse_symbol_data(cls, data: dict, symbol_separator: str) -> Tuple[Dict, Di
         return ret, info
 
     def __init__(self, **kwargs):
-        super().__init__('wss://www.bitmex.com/realtime', **kwargs)
+        auth_api = 'wss://www.bitmex.com/realtime' if not sandbox else 'wss://testnet.bitmex.com/realtime'
+        super().__init__(f'{auth_api}/realtime', **kwargs)
```
bmoscon, line 59: looks like
        super().__init__(f'{auth_api}/realtime', **kwargs)
should be
        super().__init__(auth_api, **kwargs)

bmoscon (changes requested):

cryptofeed/exchange/bitmex.py
Pcsl: Modification added. Thank you for your help!

agijsberts (commented):
You changed `__init__` to `init`, which will cause things to fail.

Furthermore, to properly switch to the testnet, you'd also need to modify the `api` and `symbol_endpoint` class attributes accordingly.

Pcsl:
> 
> 
> You changed `__init__` to `init`, which will cause things to fail.
> 
> Furthermore, to properly switch to the testnet, you'd also need to modify the `api` and `symbol_endpoint` class attributes accordingly.

Thank you for your reply. Then unfortunately I'm afraid this is well beyond my skills. So if anyone wants to give it a try I would be really grateful.

agijsberts:
If you just change the method name from `init` to `__init__` you should have something that works. The instruments and symbols will _probably_ be identical between the real and testnet anyways.

Changing the `api` and `symbol_endpoint` attributes is not possible in the way it's currently implemented in cryptofeed: these class attributes are read from the class in general, not from any instance in particular. For this reason, it would make sense imho to implement sandboxes by inheritance rather than via an argument.

bmoscon (approved):

Pcsl:
I guess in https://github.com/bmoscon/cryptofeed/blob/3d834a15f6407bd93102592d3172314b9eb2e5a6/cryptofeed/rest/rest.py#L43

`'bitmex': Bitmex(config.bitmex),`
should be
`'bitmex': Bitmex(config.bitmex, sandbox=sandbox)`
?

PR #520, opened 2021-06-15
Authenticate private channels (USER_FILLS) for FTX

Dynamikey:
This PR lays out the groundwork to authenticate private channels for a particular websocket connection. FTX and Deribit are examples of exchanges that require authenticating a connection before a private channel can be subscribed to.

- [x] - Tested
- [x] - Changelog updated
- [x] - Tests run and pass - one test was failing (tests/integration/rest/test_rest.py::test_rest_bitmex was failing on a clean branch already)

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -8,4 +8,4 @@ Cryptofeed was originally created by Bryant Moscon, but many others have contrib
 * [Ryan Tam](https://github.com/ryantam626) - <ryantam626@gmail.com>
 * [Yoh Plala](https://github.com/yohplala) - <yoh.plala@gmail.com>
 * [Alex](https://github.com/globophobe)
-
+* [Michael Zhao](https://github.com/dynamikey)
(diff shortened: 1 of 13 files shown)


bmoscon:
I'm not opposed to the way you did this, but I don't understand why you added `authenticate` to so many places. It seems like you could just authenticate in the subscribe method? Am I missing something? Just trying to understand why you went this route

Dynamikey:
I see your point. Tbh, I thought that it would be good to have authenticate function separated as there are other exchanges which would need to be authed this way, such as Deribit. So, if Deribit needed to be authed this way, all you’d need to do is implement the authenticate function.

Your method would work too.

bmoscon (approved):

bmoscon:
makes sense, seems reasonable. Thanks! Just a fair warning, when FILLS are added to other exchanges the signature of that callback might change a little

Dynamikey:
Thanks bmoscon! Will keep that in mind.

PR #542, opened 2021-07-02
Hitbtc as a child class of Bequant

peedrr:
Not sure if you saw message on Slack. I've applied all the new features and authenticated goodness from Bequant to HitBTC as they share the same API. This was achieved by making HItBTC a child of the Bequant class.

For this one, please check over the inheritance.  I know enough to be dangerous but there may well be better ways of doing things, or something that I missed. 

It's only a few lines of code... I threw most of your good work in the bin  🧛‍♂️ 

Have updated the Bequant demo to be a combined Bequant/HitBTC one.

- [X] - Tested
- [X] - Changelog updated
- [ ] - Tests run and pass


cryptofeed/exchange/old.hitbtc.py
```diff
@@ -0,0 +1,136 @@
+'''
```
bmoscon, line 1: no reason to live the file, its in the git history

cryptofeed/standards.py
```diff
@@ -115,6 +115,7 @@ def timestamp_normalize(exchange, ts):
     },
     TRADES: {
         DYDX: 'v3_trades',
+        HITBTC: 'subscribeTrades',
```
bmoscon, line 118: Why am I seeing this here? It supported trades before, are you adding a new one and leaving the old entry?

cryptofeed/standards.py
```diff
@@ -267,7 +278,9 @@ def timestamp_normalize(exchange, ts):
     },
     FILL_OR_KILL: {
         BEQUANT: {'timeInForce': 'FOK'},
+        HITBTC: {'timeInForce': 'FOK'},
```
bmoscon, line 281: looks like this was added twice?

cryptofeed/exchange/bequant.py
```diff
@@ -74,6 +73,290 @@ def __reset(self):
         self.l2_book = {}
         self.seq_no = {}
 
+    async def _ticker(self, msg: dict, ts: float):
```
bmoscon, line 76: why does it show all this code being added? This code is already present in the repo

bmoscon (changes requested):

cryptofeed/exchange/bequant.py
peedrr: Hmm. I think I might know why. I rebased on my 'bequant' branch when I should have rebased on origin (which merged the bequant commits). If you're happy with the changes in principle, reject this one and I'll start afresh on the current master

bmoscon:
@peedrr yes I think the changes in question are fine, will be easier to confirm with out all the duplicated changes. Please restart from master and we'll get it merged

PR #547, opened 2021-07-03
HitBTC & Bitcoin.com inherit Bequant features

peedrr:
I started again from a clean tree and took the opportunity to make the same changes to bitcoincom. So this means that both exchanges now inherit:
- candles
- authentication
- order_info
- acc_transactions
- acc_balances

I created one demo which showcases these features on all 3 exchanges at once and deleted the existing Bequant demo. 

- [X] - Tested
- [X] - Changelog updated
- [] - Tests run and pass


cryptofeed/exchange/bitcoincom.py
```diff
@@ -6,134 +6,40 @@
...
-                return
-            elif msg['method'] == 'updateTrades':
-                await self._trade(data, timestamp)
-            elif msg['method'] == 'ticker':
-                await self._ticker(data, timestamp)
-            else:
-                LOG.warning("%s: Invalid message type %s", self.id, msg)
-        else:
-            LOG.warning("%s: Invalid message type %s", self.id, msg)
+        super().__init__(candle_interval='1m')
```
bmoscon, line 44: I'm not sure you need both supers here, I'm pretty sure you just need

`super().__init__(urls, candle_interval=candle_interval, **kwargs)`

Also, candle_interval should be a kwarg in the constructor here

cryptofeed/exchange/hitbtc.py
```diff
@@ -6,131 +6,41 @@
...
-                            "symbol": pair
-                        },
-                        "id": conn.uuid
-                    }))
+        urls = {
+            'market': 'wss://api.hitbtc.com/api/2/ws/public',
+            'trading': 'wss://api.hitbtc.com/api/2/ws/trading',
+            'account': 'wss://api.hitbtc.com/api/2/ws/account',
+        }
+        super().__init__(candle_interval='1m')
```
bmoscon, line 45: see previous comments on bitcoincom

cryptofeed/standards.py
```diff
@@ -116,6 +116,7 @@ def timestamp_normalize(exchange, ts):
     },
     TRADES: {
         DYDX: 'v3_trades',
+        HITBTC: 'subscribeTrades',
```
bmoscon, line 119: this looks like you've duplicated it. I can see below on line 122 that its already defined. Please remove this

bmoscon (changes requested):
see previous comments

cryptofeed/exchange/bitcoincom.py
peedrr: I'm going to attempt to reply like I know what I'm talking about... ;)

The second super has Bequant as its subclass argument. This means it is calling the __init__ method from the class _above_ Bequant, i.e. Feed. And this is necessary because the urls need to be passed to Feed, not to Bequant.

I think I probably shouldn't have also passed urls to Bequant's __init__. I can remove these. But they are having no effect anyway.

bmoscon: the bequant class already calls super on the parent class (Feed). Look at okex or any other the others that subclass. you don't need to call init twice.

peedrr: OK, thanks for the pointer. I'll take a look tomorrow

peedrr: Yep, of course you were right. I've reworked it.

Re. candle_interval, I removed it altogether as this all gets inherited from parent and it was creating multiple assignments.

cryptofeed/standards.py
peedrr: I missed this one. Now fixed with latest push

bmoscon (approved):

PR #576, opened 2021-07-24
feeds http_proxy

jinusean:

- [x] - Tested
- [x] - Changelog updated
- [x] - Tests run and pass

Enables a `http_proxy` param in Feed that is used by HttpProxy/HttpPoll when making GET calls.


jinusean:
Should aid in overcoming issues described in https://github.com/bmoscon/cryptostore/issues/156

bmoscon (changes requested):
it looks like you've added a new library requirement, YARL. I'm not opposed to adding libraries if they are used, but if you are using it solely for type hinting - is it really necessary?

cryptofeed/feed.py
```diff
@@ -9,6 +9,7 @@
 import logging
 import os
 from typing import Dict, Tuple, Callable, Union, List
+from aiohttp.typedefs import StrOrURL
```
jinusean, line 12: Replaced YARL with aiohttp's typedef

cryptofeed/connection.py
```diff
@@ -15,7 +15,7 @@
 import websockets
 import requests
 from yapic import json as json_parser
-
+from aiohttp.typedefs import StrOrURL
```
jinusean, line 18: Replaced YARL with aiohttp's typedef

jinusean:
YARL was used by aiohttp but using the latter's type definitions is probably better.

Thanks!

bmoscon (approved):

PR #580, opened 2021-07-28
Add support for 'concurrent_http' requests in Binance feeds

jinusean:
- Allow `concurrent_http` requests in Binance feeds: L2_BOOK, and OPEN_INTEREST
- Created HTTPConcurrentPoll connection type for concurrent polling.
- Updated `examples/demo_proxy.py` -> `examples/demo_concurrent_proxy.py`

Please see output in `examples/demo_concurrent_proxy.py` to see performance improvements (no proxy necessary).

- [x] - Tested
- [x] - Changelog updated
- [x] - Tests run and pass


cryptofeed/connection.py
```diff
@@ -5,19 +5,20 @@
 associated with this software.
 '''
 import asyncio
-from decimal import Decimal
+import aiohttp
+import requests
```
bmoscon, line 9: im not sure why these were re-ordered. I like the imports to be in blocks of std library imports, followed by a block of external, 3rd party library imports, followed by a block of internal library imports. Please undo/correct these changes here

cryptofeed/connection.py
```diff
@@ -164,28 +165,61 @@ def __init__(self, address: Union[List, str], conn_id: str, delay: float = 60, s
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._queue: Optional[Queue] = None
+
+    async def _poll_address(self, address: str, header=None):
+        try:
+            while True:
+                data = await self._read_address(address, header)
+                await asyncio.gather(self._queue.put(data), asyncio.sleep(self.sleep))
```
bmoscon, line 204: im not sure why you are using gather here. You have already spawned a new task to run `_poll_address`, why create more tasks to put the data on the queue and then wait? cant you just `await` the queue.put and then `await` the sleep?

bmoscon (changes requested):
see comments

cryptofeed/connection.py
jinusean: The rationale was to keep the callback time more uniform to the user specified sleep time. For example, if sleep was set to 1 second, and a request takes 0.2 seconds, each callback would have a latency of 1.2 seconds. Whereas by using gather the latency would more closely resemble specified sleep time. 

Please let me know if you're not a fan of this approach and I will change the implementation.

cryptofeed/connection.py
jinusean: It must've been my IDE since it does all the importing for me automatically. I'll keep your preferences in mind for future pull requests.

bmoscon (approved):

PR #599, opened 2021-08-09
Rest rework v2

bmoscon:


bmoscon:
@Dynamikey - sorry, I had to remove the FTX REST authenticated channels. They were done in a significantly different manner than the other exchanges, and since I cannot test against FTX, I would have had very little confidence that I'd have reworked them correctly. I can at some point test against FTXUS, but it will be a while before I'm able to try to re-add them. If you want to have a go at it, the diff is here (so you can see what I remove all in one place). It may be easy for you to tweak them and re-integrate them.

Issue #604, opened 2021-08-11
incorrect OrderBook for binance


tristan-murfitt-elw:
**Describe the bug**
Over time, internal binance orderbook can be incorrect because we don't regularly poll the snapshot.
In comparison, coinbase WS feed automatically supplies regular snaps (and doesn't truncate the book!)

Tagging @jinusean because their recent PR https://github.com/bmoscon/cryptofeed/pull/580 touched a lot of this file and I'm sure they can help confirm!

**More detail**
When fetching snapshot we truncate depth to 1000 (instead of the max of 5k, `self.valid_depths[-1]`):
https://github.com/bmoscon/cryptofeed/blob/b367c0fdfcc5e4f0be6068c5d51a421cf4b4782b/cryptofeed/exchanges/binance.py#L259

A new snapshot only gets fetched when a message is dropped (missing seq num):
https://github.com/bmoscon/cryptofeed/blob/b367c0fdfcc5e4f0be6068c5d51a421cf4b4782b/cryptofeed/exchanges/binance.py#L253

I think this means, after running for a while, the internal OB could be outdated with the real world. The initial 1000 levels are recorded, and we store deltas at all(?) levels, but we will never know about unchanged levels outside of those initial 1k levels. This is a problem if the market moves in one direction over time, and the real book has an unchanged (old) limit order but our internal OB doesn't.

Do you both agree?
If so, @bmoscon do you think any other exchanges suffer from this problem?

**Solution**
- Increase snap depth to 5k to reduce the risk of this happening.
- Periodically fetch a new snapshot from the API (eg. every 30s)

**Cryptofeed Version**
latest

tristan-murfitt-elw:
Binance websocket docs ["How to manage a local order book correctly"](https://binance-docs.github.io/apidocs/spot/en/#diff-depth-stream)

Surprises me that they don't mention this potential issue (and tell you to regularly refresh snapshot)

bmoscon:
@tristan-murfitt-elw - their docs indicate that we should use 1000 as the depth, are you sure that the websocket is providing book updates for all 5000 levels?

tristan-murfitt-elw:
@bmoscon there are certainly more than 1000 levels in the binance order books so we should use the maximum available (as is done for other exchanges).

To test this, I set depth to 5 and watched the internal OB get larger and larger to 2000+ levels as the updates came in.

[...]

PR #606, opened 2021-08-15
binance bug experiment

tristan-murfitt-elw:
proof of internal orderbook inconsistency

### DIFF
--- a/cryptofeed/exchanges/binance.py
+++ b/cryptofeed/exchanges/binance.py
@@ -256,7 +256,7 @@ class Binance(Feed, BinanceRestMixin):
         return skip_update, forced, current_match
 
     async def _snapshot(self, pair: str) -> None:
-        max_depth = self.max_depth if self.max_depth else self.valid_depths[-1]
+        max_depth = self.max_depth if self.max_depth else 20
         if max_depth not in self.valid_depths:
             for d in self.valid_depths:
                 if d > max_depth:
--- a/examples/demo_binance_depth_intervals.py
+++ b/examples/demo_binance_depth_intervals.py
@@ -1,10 +1,22 @@
 from time import time
 
+from decimal import Decimal
 from cryptofeed import FeedHandler
 from cryptofeed.defines import L2_BOOK
 from cryptofeed.exchanges.binance import Binance
 from cryptofeed.exchanges.binance_futures import BinanceFutures
 from cryptofeed.exchanges import BinanceDelivery
+
+
+# Ask limit orders I manually placed in the book (20+ levels away from mid) prior to starting cryptofeed
+# These should ALL be in the internal OB if market moves upwards
+# Some may not be if the price level gets updated while cf is running (eg. by another participant)
+ASKS_TO_LOOK_FOR = [
+    (Decimal('195.66'), Decimal('0.06')),
+    (Decimal('195.67'), Decimal('0.06')),  # added this one AFTER starting cf (will be in the book)
+    (Decimal('195.68'), Decimal('0.06')),
+    (Decimal('195.69'), Decimal('0.06')),
+]
 
 
 def timer(interval):
@@ -15,8 +27,23 @@ def timer(interval):
         now = time()
 
         delta = (now - then) * 1000
-        print(
-            f'{f"{feed}":<16} {f"[depth@{interval}]":<14} {int(delta):>4}ms ago, Timestamp: {timestamp}, Receipt: {receipt_timestamp}, Snapshot: {book}')
+
+        asks = book['ask'].items()[:20]
+
+        asks_debug = [f'{px:.2f}|{sz:.3f}' for px, sz in asks]
+
+        print(f'\nSnap at {timestamp} (truncated book from {len(book["ask"])} to {len(asks)}):')
+        print(', '.join(asks_debug))
+
+        max_px_in_book = asks[-1][0]
+
+        for px, sz in ASKS_TO_LOOK_FOR:
+            if max_px_in_book < px:
+                continue
+            else:
+                if px not in [p for p, _ in asks]:
+                    print(f'ERROR. {px:.2f} not in book but should be!!')
+
         then = now
 
     return abook
@@ -26,22 +53,9 @@ def main():
     f = FeedHandler()
     # intervals = '100ms', '1000ms'
     f.add_feed(Binance(depth_interval='1000ms',
-                       max_depth=1,
-                       symbols=Binance.info()['symbols'][:1],
+                       symbols=['DASH-BUSD'],
                        channels=[L2_BOOK],
                        callbacks={L2_BOOK: timer('1000ms')}))
-
-    # default interval is 100ms, '100ms', '250ms', '500ms'
-    f.add_feed(BinanceDelivery(max_depth=1,
-                               symbols=BinanceDelivery.info()['symbols'][:1],
-                               channels=[L2_BOOK],
-                               callbacks={L2_BOOK: timer('100ms')}))
-    # intervals = '100ms', '250ms', '500ms'
-    f.add_feed(BinanceFutures(depth_interval='500ms',
-                              max_depth=1,
-                              symbols=BinanceFutures.info()['symbols'][:1],
-                              channels=[L2_BOOK],
-                              callbacks={L2_BOOK: timer('500ms')}))
     f.run()
 
 


villekuosmanen:
Not sure if I should have created a new branch or used this one Tristan made earlier...

In any case, this PR fixes the bug reported at https://github.com/bmoscon/cryptofeed/issues/604. Because the bug is fairly complex and we have to maintain a buffer of all messages we receive while waiting for the REST API response, the code here is a bit complex which I'm not a fan of. However I don't see an obvious way to simplify it.

I tested the code with Tristan's test script and it works. The test is not fully automatic (and due to to architecture of the exchange connectors it would be very difficult to write an automated test for this. I'll take another look at it later before opening it up for review.

cryptofeed/exchanges/binance.py
```diff
@@ -66,6 +68,13 @@ def _parse_symbol_data(cls, data: dict) -> Tuple[Dict, Dict]:
             info['instrument_type'][s.normalized] = stype
         return ret, info
 
+    def _max_msgs_since_snapshot(self) -> int:
```
villekuosmanen, line 71: This function decides when we should re-fetch the snapshot.

Currently it is based on the number of messages received. I'm not actually sure if that's a good idea since number of received messages != book updates. Perhaps we'd be better off on basing this on a timer instead?

cryptofeed/exchanges/binance.py
```diff
@@ -261,26 +272,107 @@ async def _snapshot(self, pair: str) -> None:
...
         resp = json.loads(resp, parse_float=Decimal)
+        return resp
+
+    async def _reset_snapshot(self, pair: str) -> None:
+        resp = await self._fetch_snapshot(pair)
 
         std_pair = self.exchange_symbol_to_std_symbol(pair)
         self.last_update_id[std_pair] = resp['lastUpdateId']
-        self._l2_book[std_pair] = OrderBook(self.id, std_pair, max_depth=self.max_depth, bids={Decimal(u[0]): Decimal(u[1]) for u in resp['bids']}, asks={Decimal(u[0]): Decimal(u[1]) for u in resp['asks']})
+        self._l2_book[std_pair] = OrderBook(self.id, std_pair, bids={Decimal(u[0]): Decimal(u[1]) for u in resp['bids']}, asks={Decimal(u[0]): Decimal(u[1]) for u in resp['asks']})
```
villekuosmanen, line 282: I don't think we want to set `max_depth` on the order book. That will always truncate the book for that size, which gets rid of useful updates higher up in the book and can contribute to the order book getting out of date.

bmoscon: max depth doesnt truncate, unless you set truncate to True, which nothing does. The data is still there and can move up if necessary

cryptofeed/exchanges/binance.py
```diff
@@ -261,26 +272,107 @@ async def _snapshot(self, pair: str) -> None:
...
+        resp = await self._fetch_snapshot(pair)
+        std_pair = self.exchange_symbol_to_std_symbol(pair)
+
+        if std_pair in self._book_buffer or std_pair not in self._fetch_snapshop_buffer:
+            # already fetching / fetched a snapshot, so ignore this message
+            self._fetch_snapshop_buffer.pop(std_pair, None)
             return
 
+        while resp['lastUpdateId'] > self.last_update_id[std_pair]:
+            # Snapshot came before we received the corresponding WS update. Wait until we receive that.
```
villekuosmanen, line 330: If the snapshot is processed before the web socket deltas have come in, the snapshot could be ahead of the local book which we don't allow.

cryptofeed/exchanges/binance.py
```diff
@@ -321,18 +429,34 @@ async def _book(self, msg: dict, pair: str, timestamp: float) -> None:
         """
         book_args = (msg, pair, timestamp)
         std_pair = self.exchange_symbol_to_std_symbol(pair)
+        is_expired_snapshot = std_pair in self._msgs_since_snapshot and self._msgs_since_snapshot[std_pair] >= self._max_msgs_since_snapshot()
 
         if not self.concurrent_http:
```
villekuosmanen, line 434: I was a bit confused on this one. Why would we not want to use concurrent http?

cryptofeed/exchanges/binance.py
```diff
@@ -261,26 +272,107 @@ async def _snapshot(self, pair: str) -> None:
...
+        # Reset messages since snapshot and delete buffer
+        self._msgs_since_snapshot[std_pair] = 0
+        self._fetch_snapshop_buffer.pop(std_pair, None)
+
+        ts = time.time()
+        await self.book_callback(L2_BOOK, self._l2_book[std_pair], ts, timestamp=ts, delta=delta, sequence_number=self.last_update_id[std_pair])
+
+
+    async def _refresh_snapshot_with_buffer(self, pair: str) -> None:
+        resp = await self._fetch_snapshot(pair)
```
hk-yuan, line 321: `ts = time.time()` here for receipt_timestamp (non-named) into `book_callback` below because might be waiting a long time in sleeps during `while resp['lastUpdateId'] > self.last_update_id[std_pair]:`

cryptofeed/exchanges/binance.py
```diff
@@ -261,26 +272,107 @@ async def _snapshot(self, pair: str) -> None:
...
         std_pair = self.exchange_symbol_to_std_symbol(pair)
-        skip_update, current_match = self._check_update_id(std_pair, msg)
-        if current_match:
-            # Current msg.final_update_id == self.last_update_id[pair] which is the snapshot's update id
-            return await self.book_callback(L2_BOOK, self._l2_book[std_pair], timestamp, raw=msg, sequence_number=self.last_update_id[std_pair], timestamp=self.timestamp_normalize(msg['E']))
+        self.last_update_id[std_pair] = resp['lastUpdateId']
 
-        if skip_update:
+        delta = {BID: [], ASK: []}
+        for key, side in [('bids', BID) , ('asks', ASK)]:
```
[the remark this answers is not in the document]
hk-yuan: maybe this func `_append_msg_to_internal_book` https://github.com/bmoscon/cryptofeed/pull/606/files#diff-05706512ccd2e1413e7099f1923bc66662b6463d4b87d353de2342a6c01d92c8R375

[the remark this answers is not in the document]
villekuosmanen: The function can't be reused here but can be one the one below (`_refresh_snapshot_with_buffer`)

cryptofeed/exchanges/binance.py
villekuosmanen: done

villekuosmanen:
This PR is now ready for review.

It adds a new method `refresh_snapshot` which is similar to the old snapshot method, except that it doesn't fully reset the local book but rather adds new prices into it. Two versions exist: a regular one for asynch http, and a buffered one for concurrent http (which is a bit complicated sadly).

The refresh is triggered based on time since last message. The time depends on the order book depth: smaller books need to refresh more frequently. The time increase is not linear but tapers off as order books increase: order book of 10 refreshes every 10 seconds, while a book of 5000 (default) refreshes every ~15 mins 47 secs. This should be often enough to reduce the risk of the top of the the book going out of date to minimum, while being rarely enough that is shouldn't cause too much network load.

A couple of pointers:

1. Would you like to add a parameter to turn the order book refresh off?
2. Should we add some randomisation on the fetch timer? Currently, most pairs would trigger a refresh at the same time which could cause an unexpected load. Adding some variation would distribute the API requests over time.

Any other thoughts?

bmoscon:
@villekuosmanen - either randomization in regards to the timing (say refresh at a random time interval within a pre-determine range) or put in some code to detect and handle 429s on snapshots appropriately. If only doing one, handling 429 is probably the best solution (with a random wait/backoff), but doing both together would be ideal. Depends on how much time you want to spend on it.

bmoscon (changes requested):
see previous comment about handling 429s or randomizing the refresh slightly within a range.

nirvana-msu:
Refreshing on a timer is far from the best strategy though, isn't it? We have an almost deterministic problem here. We know the true state of the order book between the bottom-most and top-most levels returned by the previous snapshot. So long as the price stays comfortably within that range, we never even need to refresh again - we know the true order book here. It's only when the price starts to get close to either lower or the upper bound - that's when we start risking that the requested `max_depth` levels could fall outside those bounds where we don't know the true state.

I would say the ideal logic should be based on how far the current price is from those bounds, and requested `max_depth`. When the price gets close, we should refresh. 

If we value integrity (correctness) of the data over timeliness, we could take it a step further and do not push a callback to the user unless we know we have enough levels fetched in the last snapshot (and instead wait for the new snapshot if we don't have enough known levels). That way we could guarantee data integrity, at the expense of potential delays. And to avoid those delays the code would need to be smart enough to know when to request a new snapshot in advance (but it can still happen unexpectedly on a large price gap moves).

Also, using the above strategy instead of a timer should help with rate limits. If the prices are not all moving in tandem, then refresh for each symbol would happen fairly independently from each other.

nirvana-msu:
>  The time increase is not linear but tapers off as order books increase: order book of 10 refreshes every 10 seconds, while a book of 5000 (default) refreshes every ~15 mins 47 secs. 

This logic probably makes it infeasible to pull all Binance symbols when setting `max_depth=10`. i.e. it would simply be rate-limited, trying to pull a snapshot of every symbol every 10 seconds.

Expanding on my comment above, `max_depth=10` that the user may request does not imply that you need to maintain only a small order book internally. It may be better to main a large order book internally instead. The upside of choosing a large size for internal order book is that we'll need to refresh it less frequently (only when a price moves close to a boundary, as descrived above) - which would help with rate limits, especially when nubmer of symbols subscribed to is large. The obvious downside is performance - it is more time-consuming to maintain a larger order book. Ideally the logic should find that fine balance between performance and rate limits - order book should be large enough to avoid hitting the rate limits, but not too large as it would unnecessarily affect performance.

nirvana-msu:
And just to repeat my comments in https://github.com/bmoscon/cryptofeed/issues/604, no matter how often you refresh, if you only pull exactly `max_depth` levels in a REST snapshot, you'll end up having fewer levels than `max_depth` when levels are removed from the order book (or when the price moves even a little bit). My comment in the linked issue has an example you can run - it only takes seconds to run into this problem (actually, milliseconds if you change depth_interval to 100ms). The size of the snapshot maintained internally should be much larger - see my previous comment on the considerations (i.e. it being a balance between rate limits and performance).

bmoscon:
also @villekuosmanen - there were some issues in the book handling in Binance that I had to fix, which has lead to a merge conflict. Its not explicit anywhere, but delta must be set to None on the first book callback from all exchanges. Binance was not honoring this, so I changed it so the snapshot would trigger a book update. If you want to change how this works, thats fine, but keep in mind the very first snapshot after a reset or upon subscribe needs to have delta set to None.

bmoscon:
@tristan-murfitt-elw  / @villekuosmanen - just wanted to get your idea of a timeframe for resolving these last few issues on this. I'm going to do a release soon so please let me know

tristan-murfitt-elw:
@bmoscon i'll spend some time reviewing this week, not trivial to solve this in a tidy way. Poor effort from Binance on this one IMO

> So long as the price stays comfortably within that range, we never even need to refresh again - we know the true order book here. It's only when the price starts to get close to either lower or the upper bound - that's when we start risking that the requested max_depth levels could fall outside those bounds where we don't know the true state.

@nirvana-msu I agree with all your points on performance / rate limits etc, but this does neglect the users who want an accurate book at less-competitive limit levels too. This would still result in incorrect data at the ends of the book (eg. one trade and 5001th level is no longer there, so your book only has 4999 levels) as you pointed out with the `max_depth` problem.

This is clearly a limitation of Binance, there is no easy solution. Maybe some combination of the below:

1. timer with randomness to periodically fetch snapshots 
2. never fetch a snapshot <100 levels for Binance only (due to these problems), always fetch snapshot > `max_depth`
3. better name for `max_depth` which clearly states that it's used for generating synthetic book updates to keep end user at exactly this depth, and that it should very very rarely be used for performance reasons (esp for high values)
4. remove the `concurrent http` flag because it should be the default and only behaviour of the lib (reduce complexity and confusion here!)

bmoscon:
@tristan-murfitt-elw if you remove the concurrent http flag (and thus force it to always parallelize the snapshot requests) you'll pretty much have to add the checks for 429s and random backoffs

nirvana-msu:
> @nirvana-msu I agree with all your points on performance / rate limits etc, but this does neglect the users who want an accurate book at less-competitive limit levels too. This would still result in incorrect data at the ends of the book (eg. one trade and 5001th level is no longer there, so your book only has 4999 levels) as you pointed out with the max_depth problem.

@tristan-murfitt-elw Right, that's a good point. To be exact, what I described above implies that getting accurate book at 5000 levels would require re-fetching snapshots almost all the time. So it's infeasible given rate limits (and also very slow). It's essentially infeasible whatever approach we take. But yeah one thing I missed out in the logic I described would be some way to throttle REST requests if the logic wants to make them too often.

On the note of throttling, and also related to this comment
> if you remove the concurrent http flag (and thus force it to always parallelize the snapshot requests) you'll pretty much have to add the checks for 429s and random backoffs

it might be better to be pro-active with throttling rather than re-active - i.e, respect rate limits when firing requests. Something like [aiolimiter](https://github.com/mjpieters/aiolimiter) is a fairly neat solution to this  (though that exact implementation uses leaky bucket algorithm which may not be exactly how Binance enforces the limit). `aiolimiter` also supports requests of varying capacity, which is exactly what's needed here.

bmoscon:
your comment assumes that they only interact with the exchange via this single library, which is unlikely to be true. Definitely need to handle 429s if we're going to constantly refresh the snapshots

nirvana-msu:
Yes, I wasn't clear - you definitely still need to handle 429s. But it might be slightly more efficient to do both - enforce internal rate limiting, and handle any 429 on top. You'll hit fewer of those 429 that way.

villekuosmanen:
Based on Binance's docs, the API rate limit is 1200 request weight / minute. In terms of 5000 depth snapshots, this is 24 snapshots in a minute. If we fetched such a snapshot every 15 minutes for 100 symbols (can you even sub to that many in a single instance?) it would cause a load of 333 request weight / minute. It's not unreasonable that we would exceed the limits even with random variation in the snapshot timer.

Handling 429 errors does add complexity (again!) which will make the PR even bigger which is going to be annoying. We will probably need to do that at some point, but in the interest of time, we could probably avoid doing it for now by:

- Adding a `refresh_snapshot` parameter that enables or disables this behaviour based on user preference, and defaulting it to False
- Documenting the use of the above parameter, and how when it's enabled you run a risk of breaching rate limits when subscribing to a large number of symbols
- Regardless of `max_depth`, fetch snapshots of at least 100 orders.

Any thoughts?

bmoscon:
@villekuosmanen sounds reasonable.

villekuosmanen:
I have now added random variation into the snapshot fetch time (it will be between 0.5x and 1.5x the expected time). In addition, shapshots will always be fetched at a minimum of 100 depth. at depths <= 100 we will refresh a snapshot every 30 - 90 seconds. The request weight of an order book depth of 100 is 1, which means there's essentially no risk of hitting the 1200 request weight limit per minute with them, even if you use other libraries to query the API at the same time.

I still added a parameter `refresh_snapshot` to turn this functionality off by default since at large order book depths it's possible to hit the rate limit when subscribed to many symbols at the same time.

bmoscon:
@villekuosmanen - is this good to merge now?

villekuosmanen:
@bmoscon it's good on my end. Would you mind taking a look as well before merging?

cryptofeed/exchanges/binance.py
```diff
@@ -287,36 +322,140 @@ async def _snapshot(self, pair: str) -> None:
...
         self._l2_book[std_pair] = OrderBook(self.id, std_pair, max_depth=self.max_depth, bids={Decimal(u[0]): Decimal(u[1]) for u in resp['bids']}, asks={Decimal(u[0]): Decimal(u[1]) for u in resp['asks']})
+        self._l2_book[std_pair].sequence_number = resp['lastUpdateId']
+        self._l2_book[std_pair].raw = resp
+        
+        if self.refresh_snapshot:
+            # Delete buffer and reset time since snapshot update
+            self._next_snapshot_time[std_pair] = receipt_timestamp + self._next_snapshot_refresh_interval()
+            self._fetch_snapshop_buffer.pop(std_pair, None)
         ts = self.timestamp_normalize(resp['E']) if 'E' in resp else None
-        await self.book_callback(L2_BOOK, self._l2_book[std_pair], time.time(), raw=resp, timestamp=ts, sequence_number=resp['lastUpdateId'])
```
bmoscon, line 295: setting them above wont work, book_callback will reset them to None. They need to be passed in like they were previously

cryptofeed/exchanges/binance.py
```diff
@@ -287,36 +322,140 @@ async def _snapshot(self, pair: str) -> None:
...
+                        delta[side].append((price, amount))
+        self._l2_book[std_pair].sequence_number = resp['lastUpdateId']
+        
+        if self.refresh_snapshot:
+            # Delete buffer and reset time since snapshot update
+            self._next_snapshot_time[std_pair] = receipt_timestamp + self._next_snapshot_refresh_interval()
+            self._fetch_snapshop_buffer.pop(std_pair, None)
+
+        await self.book_callback(L2_BOOK, self._l2_book[std_pair], receipt_timestamp, delta=delta, sequence_number=self.last_update_id[std_pair])
+        self._l2_book[std_pair].delta = None  # Set delta as None to allow older messages to be skipped correctly
```
bmoscon, line 376: I'm not sure that setting it after the callback will do anything. What is the purpose of this? It will just be reset at the next callback

bmoscon (changes requested):

bmoscon:
@villekuosmanen - very close just a few last comments / questions

cryptofeed/exchanges/binance.py
villekuosmanen: Makes sense

cryptofeed/exchanges/binance.py
villekuosmanen: This method is used in refreshing a snapshot asynchronously. After we have processed this , the next web socket message could have a lastUpdateId <= this ID. This is handled in the `_check_update_id` method, which seemed to have changed in the last few weeks to remove a clause that allowed us to ignore such messages. Without the clause, this case would result in the order book being reset as the system would think it dropped a message. Setting the `delta` as None allows us to fall on the following if statement:
```
if self._l2_book[std_pair].delta is None and msg['u'] <= self.last_update_id[std_pair]:
    return True
```

Tbh it's a hack and probably a better idea would be to add a clause such as:
```
if msg['U'] <= self.last_update_id[std_pair] and msg['u'] <= self.last_update_id[std_pair]:
    return True
```

It feels like if both the start and end IDs are smaller than our last seen ID, we would be able to ignore that message safely. What do you think? I'd prefer removing this line and adding the above check instead.

cryptofeed/exchanges/binance.py
```diff
@@ -287,36 +322,140 @@ async def _snapshot(self, pair: str) -> None:
...
 
+    async def _handle_book_msg(self, msg: dict, pair: str, timestamp: float):
+        """
+        Processes 'depthUpdate' update book msg. This method should only be called if pair's l2_book exists.
+        """
+        std_pair = self.exchange_symbol_to_std_symbol(pair)
+        skip_update = self._check_update_id(std_pair, msg)
+        # if current_match:
+        #     # Current msg.final_update_id == self.last_update_id[pair] which is the snapshot's update id
+        #     return await self.book_callback(L2_BOOK, self._l2_book[std_pair], timestamp, raw=msg, sequence_number=self.last_update_id[std_pair], timestamp=self.timestamp_normalize(msg['E']))
```
villekuosmanen, line 454: Also forgot to remove this code, will do so before merging

cryptofeed/exchanges/binance.py
bmoscon: yes that would make more sense and would be easier to read/follow

villekuosmanen:
@bmoscon I've updated the PR, should be ready to merge now.

tristan-murfitt-elw:
@villekuosmanen I'll give it a look over today

bmoscon:
@tristan-murfitt-elw let me know if you're ok with these changes. if i dont hear from you in the next ~15 hours or so I'll assume you're fine with the changes and merge it (assuming I have no more comments/requests for change)

cryptofeed/exchanges/binance.py
```diff
@@ -288,36 +326,136 @@ async def _snapshot(self, pair: str) -> None:
...
+                    if amount != self._l2_book[std_pair].book[side][price]:
+                        # This should never happen
+                        self._reset()
+                        LOG.error("%s: Book amount not equal to snapshot amount after fetching a refresh snapshot, resetting book", self.id)
+                        return
+                else:
+                    # Price not in order book: add it there
+                    self._l2_book[std_pair].book[side][price] = amount
+                    if i < max_depth:
+                        delta[side].append((price, amount))
```
tristan-murfitt-elw, line 422: not sure if this is safe for users who omit `self.max_depth` and want the internal book to be as large & correct as possible

I havent tested this but pretty sure that `self.max_depth` being set will enforce the book to filter before sending callback deltas anyway? I'm thinking this should always be adding a delta...

examples/demo_binance_depth_intervals.py
```diff
@@ -0,0 +1,72 @@
...
+# Ask limit orders I manually placed in the book (20+ levels away from mid) prior to starting cryptofeed
+# These should ALL be in the internal OB if market moves upwards
+# Some may not be if the price level gets updated while cf is running (eg. by another participant)
+ASKS_TO_LOOK_FOR = [
+    (Decimal('236.8'), Decimal('0.06')),
+    (Decimal('236.9'), Decimal('0.06')),
+    (Decimal('237.0'), Decimal('0.06')),  # added this one AFTER starting cf (will be in the book)
+    (Decimal('237.1'), Decimal('0.06')),
+]
+
```
tristan-murfitt-elw, line 20: this entire demo file is now completely irrelevant and needs to be deleted

tristan-murfitt-elw (commented):
a couple of things I think are needed before merging this in

FYI - Binance have confirmed the mistake on their end but are not expecting to patch this any time soon due to the complexity on their side.

cryptofeed/exchanges/binance.py
villekuosmanen: You're right, limiting deltas to max_depth is a problem if you want the book to be as big as possible. However, `max_depth` does not filter the deltas being sent to downstream consumers. That means we do need to do this filtering somewhere, otherwise a downstream consumer with max_depth=20 would see deltas for depth=38 for instance (because we fetch a snapshot of 100 depths minimum). In fact this problem exists without the snapshots as well, because the web socket could get a book update orders higher on the order book.

I'm not immediately sure how to fix this in the most elegant way. You could either change the cause to be `if i < max_depth or not self.max_depth` but that doesn't feel very nice and only fixes the issue for snapshots, not for book updates. Alternatively you could filter the deltas in the callback if max_depth is set, but that's very inefficient and rather cumbersome. @bmoscon what do you think?

villekuosmanen: Just to show that `max_depth` is not filtering deltas: when running the demo on `master` with BTC-BUSD, I got the following update:
```
Snap at 1632908947.502313 (truncated book from 20 to 20):
42221.54|0.451, 42221.57|0.021, 42221.58|0.176, 42221.65|0.083, 42223.44|0.062, 42223.45|0.083, 42223.50|0.035, 42224.11|0.059, 42225.45|0.011, 42227.01|0.001, 42228.37|0.118, 42230.47|0.128, 42230.48|0.010, 42230.49|0.054, 42230.57|0.250, 42230.98|0.002, 42231.41|0.128, 42231.68|0.059, 42232.40|0.050, 42232.78|0.126
Deltas:	[(Decimal('42221.54000000'), Decimal('0.45091000')), (Decimal('42221.57000000'), Decimal('0.02130000')), (Decimal('42221.58000000'), Decimal('0.17597000')), (Decimal('42221.61000000'), Decimal('0E-8')), (Decimal('42234.73000000'), Decimal('0.22972000')), (Decimal('42258.50000000'), Decimal('4.05378000')), (Decimal('42305.55000000'), Decimal('0.21000000'))]
```
As you can see, the last few deltas fall outside our range. That's because the Binance web socket always sends all book updates to us, even if they were outside our requested `max_depth`, and the code that filters deltas was removed some time ago.

bmoscon: thats expected. at some point ill add delta filtering either in the cython module or in the c module. its far too slow to try and do it on the python side. so, I'd say this is expected behavior for the time being

bmoscon:
@villekuosmanen - I'd suggest on snapshot refreshes to just send the update with delta set to null. At some future point i'll finish the automatic delta construction. If you make that change and the others suggested by @tristan-murfitt-elw I'll merge this

villekuosmanen:
> I'd suggest on snapshot refreshes to just send the update with delta set to None. At some future point i'll finish the automatic delta construction. If you make that change and the others suggested by @tristan-murfitt-elw I'll merge this

I don't think that's a good idea. It means the downstream consumers will never get these deltas, which could lead to an inconsistent order book. So I think we should continue publishing deltas on snapshot refresh. I think the two solutions that are possible are:

1. Don't filter deltas by `max_depth` and just publish everything. This is "wrong" but the issue already exists in the current code and like you said it makes more sense to do this filtering on the Cython code.
2. Use the above mentioned `if i < max_depth or not self.max_depth` clause. It's not the nicest in terms of code though.

I think option 1 is the best way to go. I'll be updating the PR later today to add that. What do you think?

villekuosmanen:
@bmoscon Should be ready to merge now, can you do a final check everything is okay?

bmoscon:
@villekuosmanen - looks good to me! thanks for all your work on this. @tristan-murfitt-elw - thanks for raising and driving this as well. Much thanks to you both

nirvana-msu:
I've tried running current master, and I get two serious issues:
1) Order book I get is often inverted - bids are higher than asks. The logic clearly applies some updates incorrectly. I could post some code if needed, but it should be easily reproducible. I'm running with parameters `max_depth=10,  depth_interval='100ms', concurrent_http=False, refresh_snapshot=True`.
2) There's ~2 minutes of data missing every time a snapshot is refreshed. That's pretty bad I would say - and especially so because it is not documented. There is no fundamenral reason why a snapshot refresh should lead to lost data. It's certainly possible to write the logic in such a way that no message would be lost - you need to keep processing messages using the old snapshot right up until the point new snapshot is received (and also buffer websocket updates that we may need to apply on top of the new snapshot)

nirvana-msu:
Here's a short code snippet clearly showing both issues:

```python
import logging
import sys

from cryptofeed.defines import BINANCE, L2_BOOK
from cryptofeed.types import OrderBook

format_ = '%(asctime)s [%(threadName)-10s|pid:%(process)-5d] [%(name)s] %(levelname)s: %(message)s'
logging.basicConfig(level=logging.INFO, format=format_, handlers=[logging.StreamHandler(sys.stderr)])

logger = logging.getLogger(__name__)

last_update = None


async def l2_book(ob: OrderBook, receipt_timestamp: float):
    if ob.timestamp is None:
        logger.info("Received snapshot update.")
        return

    book = ob.book.to_dict()
    bid1 = next(iter(book['bid']))
    ask1 = next(iter(book['ask']))
    if bid1 > ask1:
        logger.info(f"Inverted order book: bid1({bid1}) > ask1({ask1}).")

    global last_update
    if last_update is not None:
        delay = ob.timestamp - last_update
        if delay > 0.2:
            logger.info(f"Missing updates - last update was {delay:.1f} seconds.")
    last_update = ob.timestamp


fh = FeedHandler()
fh.add_feed(
    BINANCE, symbols=['BTC-USDT'], channels=[L2_BOOK], callbacks={L2_BOOK: l2_book},
    max_depth=10, depth_interval='100ms', concurrent_http=False, refresh_snapshot=True,
)
fh.run()
```

Output:
```
2021-10-10 02:53:41,810 [MainThread|pid:2599557] [__main__] INFO: Received snapshot update.
2021-10-10 02:54:37,059 [MainThread|pid:2599557] [__main__] INFO: Received snapshot update.
2021-10-10 02:54:37,061 [MainThread|pid:2599557] [__main__] INFO: Missing updates - last update was 0.4 seconds.
2021-10-10 02:54:47,757 [MainThread|pid:2599557] [__main__] INFO: Missing updates - last update was 0.2 seconds.
2021-10-10 02:54:59,063 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54903.83000000).
2021-10-10 02:54:59,165 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54903.83000000).
2021-10-10 02:54:59,262 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54903.83000000).
2021-10-10 02:54:59,364 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54903.83000000).
2021-10-10 02:54:59,464 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54903.83000000).
2021-10-10 02:55:01,565 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54906.02000000).
2021-10-10 02:55:01,665 [MainThread|pid:2599557] [__main__] INFO: Inverted order book: bid1(54906.65000000) > ask1(54906.02000000).
...
```

After first snapshot things look normal, however as soon as snapshot is refreshed again (~1 minute later for depth=10) things go crazy. Firstly, we notice there are some updates missing - it my case it was 3 updates, or ~300ms worth of updates - probably the time it took to fetch the snapshot.

Shortly after, book state becomes a total mess with bids and asks completely out of line.

nirvana-msu:
And while we're on the subject, not sure if there's much benefit in pushing a callback to the user with data from the snapshot itself? It seems like that data does not carry exchange timestamp, so I actually need to filter it out in my callback (I need exchange timestamp). Websocket updates are supposed to come regularly, every 100ms (or 1000ms) - why would anyone need another message once in a while that is not even as complete as a websocket update message?

bmoscon:
@tristan-murfitt-elw / @villekuosmanen any ideas on this?

tristan-murfitt-elw:
I suggest reverting the PR and we can investigate the code in more detail. If @nirvana-msu is seeing these problems then this fix is no better than Binance's buggy book in the first place.

I will be able to look in detail next week

bmoscon:
github wont let me revert it, so I'm going to just undo the changes for this and the concurrent http manually

PR #615, opened 2021-08-31
Add authentication to Phemex user data stream

strombom:
Phemex user data stream was not authenticated and no data was received.

Fixes regression error from https://github.com/bmoscon/cryptofeed/commit/311150b8ee1b09060662e9f019c3f1d43d4e0d31 when authenticate method was added to phemex.py.

- [x] - Tested
- [x] - Changelog updated
- [x] - Tests run and pass


strombom:
USER_DATA should probably be removed instead, and BALANCE added to Phemex.

PR #616, opened 2021-08-31
Add private ORDER_INFO subscription

KunduzB:
Make it possible to subscribe private ORDER_INFO data for Bitmex.

- Include ORDER_INFO channel to Bitmex
- Add parser for order message


KunduzB:
Close

PR #631, opened 2021-09-15
fix/kraken-timestamp - convert string timestamp to a float when passing it to a callback

villekuosmanen:
In `Types.pyx`, the timestamp of a `Candle` is supposed to be a float (or None) but this is not enforced by the type system:
```
cdef class Candle:
    cdef readonly str exchange
    ...
    cdef readonly object timestamp  # None or float
    ...
```
I noticed that in the Kraken exchange code for Candles, the timestamp is not correctly converted to a float, but is rather passed as a string. This one-liner fixes the issue.

Also this is my first PR on this repo, what exactly do you expect regarding the testing, changelog and stuff? I didn't open an issue for this because it's such a small bugfix.

- [ ] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass

### DIFF
--- a/cryptofeed/exchanges/kraken.py
+++ b/cryptofeed/exchanges/kraken.py
@@ -208,7 +208,7 @@ class Kraken(Feed, KrakenRestMixin):
             Decimal(low),
             Decimal(volume),
             None,
-            start,
+            float(start),
             raw=msg
         )
         await self.callback(CANDLES, c, timestamp)


bmoscon:
@villekuosmanen - this is fine, one question is have though is if all the other data channels also have timestamp as a string? I'll merge this as-is, but let me know if you looked into this or not.

villekuosmanen:
Hey Bryant, I think the majority send out the timestamp as a float but I can't guarantee if all of them do. There's a lot of callbacks and missing the conversion is a really subtle bug that's hard to notice. I'll be going over the code over the next few weeks to see if I can find more issues like this.

PR #655, opened 2021-09-27
Add support for Binance websocket orders stream

Dynamikey:
- [x] - Tested
- [x] - Changelog updated
- [x] - Tests run and pass

### DIFF
--- a/cryptofeed/exchanges/binance.py
+++ b/cryptofeed/exchanges/binance.py
@@ -20,7 +20,7 @@ from cryptofeed.defines import ASK, BALANCES, BID, BINANCE, BUY, CANDLES, FUNDIN
 from cryptofeed.feed import Feed
 from cryptofeed.symbols import Symbol
 from cryptofeed.exchanges.mixins.binance_rest import BinanceRestMixin
-from cryptofeed.types import Trade, Ticker, Candle, Liquidation, Funding, OrderBook
+from cryptofeed.types import Trade, Ticker, Candle, Liquidation, Funding, OrderBook, OrderInfo
 
 
 LOG = logging.getLogger('feedhandler')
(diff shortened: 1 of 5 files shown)


bmoscon:
What version did you test against? This wont work on 2.0+, the callbacks only take 2 arguments, the data object and the receipt timestamp. There is an orderinfo object defined here: https://github.com/bmoscon/cryptofeed/blob/master/cryptofeed/types.pyx#L349 you can see what data it takes based on the kwargs. If its missing a field you think its important, let me know and we can add it. Similarly, see the other places order info has been implemented (one example here): https://github.com/bmoscon/cryptofeed/blob/1b8f8594db3d66486fac148d6bb769dce98d997c/cryptofeed/exchanges/ftx.py#L394

bmoscon (changes requested):
see my comment, this wont work in 2.0+

Dynamikey:
> What version did you test against? This wont work on 2.0+, the callbacks only take 2 arguments, the data object and the receipt timestamp. There is an orderinfo object defined here: https://github.com/bmoscon/cryptofeed/blob/master/cryptofeed/types.pyx#L349 you can see what data it takes based on the kwargs. If its missing a field you think its important, let me know and we can add it. Similarly, see the other places order info has been implemented (one example here):
> 
> https://github.com/bmoscon/cryptofeed/blob/1b8f8594db3d66486fac148d6bb769dce98d997c/cryptofeed/exchanges/ftx.py#L394

My fault there - I was on an old version of Cryptofeed when I originally wrote the changes. Fixed now.

cryptofeed/exchanges/binance.py
```diff
@@ -470,6 +471,57 @@ async def _account_update(self, msg: dict, timestamp: float):
...
+            self.exchange_symbol_to_std_symbol(msg['s']),
+            msg['i'],
+            BUY if msg['S'].lower() == 'buy' else SELL,
+            msg['x'],
+            LIMIT if msg['o'].lower() == 'limit' else MARKET if msg['o'].lower() == 'market' else None,
+            Decimal(msg['Z'] / Decimal(msg['z'])) if not Decimal.is_zero(Decimal(msg['z'])) else None,
+            Decimal(msg['q']),
+            Decimal(msg['q']) - Decimal(msg['z']),
+            self.timestamp_normalize(msg['E']),
+        )
```
bmoscon, line 522: only thing missing would be to add the raw kwarg and store the raw message, but I can add this

bmoscon (approved):

cryptofeed/exchanges/binance.py
Dynamikey: Thanks Bryant!

PR #658, opened 2021-09-28
Fix: assert that timestamp is a float, not Decimal

mCodingLLC:
A recently added assertion in types.pyx asserts that a timestamp should be None or a Decimal, causing runtime crashes when calling to_dict on an OrderBook. This appears to be a simple typo, which has been fixed.

### DIFF
--- a/cryptofeed/types.pyx
+++ b/cryptofeed/types.pyx
@@ -316,7 +316,7 @@ cdef class OrderBook:
     def to_dict(self, delta=False, as_type=None) -> dict:
         assert self.sequence_number is None or isinstance(self.sequence_number, int)
         assert self.checksum is None or isinstance(self.checksum, (str, int))
-        assert self.timestamp is None or isinstance(self.timestamp, Decimal)
+        assert self.timestamp is None or isinstance(self.timestamp, float)
 
         def helper(x):
             if isinstance(x, dict):


bmoscon (approved):

bmoscon:
@mCodingLLC - looks good thanks. You only saw this error when you built with the type checking enabled, right? It should be disabled by default

mCodingLLC:
No I was not specifically intending to build with type checking enabled. I see this 
```python
# comment out line below to enable type checking in cython code (via assertions)
os.environ['CFLAGS'] = '-DCYTHON_WITHOUT_ASSERTIONS'
```
in the code, which I did not comment out. Is setting this environment variable supposed to be picked up by the cython compiler? Perhaps that works on some systems, but apparently not on mine. It appears that the Extension class has support for this: https://setuptools.pypa.io/en/latest/deprecated/distutils/apiref.html?highlight=Extension%20class#distutils.core.Extension. Here is an example from the cython docs: https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#configuring-the-c-build

bmoscon:
@mCodingLLC how did you install / build? Did you just install from pypi (a wheel) or did you build from source? What OS are you on?

bmoscon:
you'll notice I don't use the extensions object because I cannot get it to put the compiled binary into the correct module (I want it under cryptofeed, not at the root level) so I use cythonize which does do this correctly.

mCodingLLC:
I just did a `pip install --upgrade cryptofeed` inside my venv, which already had an older version installed. This was on a Win 10 machine. I don't think the `--upgrade` made a difference though, I jsut did a `pip install --verbose --force-reinstall cryptofeed` and here are the build commands it decided to use (notice no appearance of `CYTHON_WITHOUT_ASSERTIONS`):
```
 C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\bin\HostX86\x64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\Use
rs\James\PycharmProjects\CryptoFeedTest\venv\include -IC:\Program Files\Python39\include -IC:\Program Files\Python39\include -IC:\Program Files (x86)\Microsoft
 Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\ATLMFC\include -IC:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30
037\include -IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um -IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt -IC:\Program Fil
es (x86)\Windows Kits\10\include\10.0.19041.0\shared -IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\um -IC:\Program Files (x86)\Windows Kits\1
0\include\10.0.19041.0\winrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\cppwinrt /Tccryptofeed/types.c /Fobuild\temp.win-amd64-3.9\Release
\cryptofeed/types.obj
  types.c
  C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\bin\HostX86\x64\link.exe /nologo /INCREMENTAL:NO /LTCG /DLL /MANIFE
ST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:C:\Users\James\PycharmProjects\CryptoFeedTest\venv\libs /LIBPATH:C:\Program Files\Python39\libs /LIBPATH:C:\Program File
s\Python39 /LIBPATH:C:\Users\James\PycharmProjects\CryptoFeedTest\venv\PCbuild\amd64 /LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\
Tools\MSVC\14.29.30037\ATLMFC\lib\x64 /LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.29.30037\lib\x64 /LIBPATH:C:\Pr
ogram Files (x86)\Windows Kits\NETFXSDK\4.8\lib\um\x64 /LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.19041.0\ucrt\x64 /LIBPATH:C:\Program Files (x
86)\Windows Kits\10\lib\10.0.19041.0\um\x64 /EXPORT:PyInit_types build\temp.win-amd64-3.9\Release\cryptofeed/types.obj /OUT:build\lib.win-amd64-3.9\cryptofee
d\types.cp39-win_amd64.pyd /IMPLIB:build\temp.win-amd64-3.9\Release\cryptofeed\types.cp39-win_amd64.lib
```

Issue #671, opened 2021-10-02
Race Condition in Binance L2_BOOK feed/snapshot logic when concurrent_http is True and feed is restarted


nirvana-msu:
**Describe the bug**
There is a race condition bug within snapshot handling logic for Binance feed when using ` concurrent_http=True `, and feed is getting restarted for whatever reason. It leads either to `KeyError` exceptions, or possibly even to incorrect order book state in certain cases.

**To Reproduce**
You can run the following script to reproduce. I've tested it against the current release, `2.0.1`. I've reduced timeout settings to make it trigger more often - with these settings I can see exceptions ~ once every minute.

```python
import asyncio
import logging
import sys
import time

from cryptofeed import FeedHandler
from cryptofeed.defines import BINANCE, L2_BOOK
from cryptofeed.types import OrderBook

format_ = '%(asctime)s [%(threadName)-10s|pid:%(process)-5d] [%(name)s] %(levelname)s: %(message)s'
logging.basicConfig(level=logging.INFO, format=format_, handlers=[logging.StreamHandler(sys.stderr)])

logger = logging.getLogger(__name__)

counter = 0


async def l2_book(ob: OrderBook, receipt_timestamp: float):
    global counter
    if counter % 2 == 0:
        logger.info("Sleeping for 5 seconds")
        time.sleep(5)
        await asyncio.sleep(1)
        logger.info("Finished sleeping")
    counter += 1


fh = FeedHandler()
fh.add_feed(
    BINANCE, symbols=['BTC-USDT'], channels=[L2_BOOK], callbacks={L2_BOOK: l2_book},
    max_depth=10, depth_interval='1000ms', concurrent_http=True, timeout=4, timeout_interval=1,
)
fh.run()
```

Sample output:

```
2021-10-02 01:57:53,257 [MainThread|pid:1539813] [__main__] INFO: Sleeping for 5 seconds
2021-10-02 01:57:58,263 : WARNING : BINANCE.ws.1: received no messages within timeout, restarting connection
2021-10-02 01:57:59,262 [MainThread|pid:1539813] [__main__] INFO: Finished sleeping
2021-10-02 01:57:59,264 [MainThread|pid:1539813] [__main__] INFO: Sleeping for 5 seconds
2021-10-02 01:58:05,271 [MainThread|pid:1539813] [__main__] INFO: Finished sleeping
[...]

PR #673, opened 2021-10-03
Fix Binance snapshot race condition

jinusean:
Fixes https://github.com/bmoscon/cryptofeed/issues/671

- [X] - Tested
- [X] - Changelog updated
- [X] - Tests run and pass
- [X] - Contributors file updated (optional)

### DIFF
--- a/cryptofeed/exchanges/binance.py
+++ b/cryptofeed/exchanges/binance.py
@@ -363,10 +363,12 @@ class Binance(Feed, BinanceRestMixin):
 
         async def _concurrent_snapshot():
             await self._snapshot(pair)
-            while len(self._book_buffer[std_pair]) > 0:
+            while self._book_buffer and  std_pair in self._l2_book and len(self._book_buffer[std_pair]) > 0:
                 book_args = self._book_buffer[std_pair].popleft()
                 await self._handle_book_msg(*book_args)
-            del self._book_buffer[std_pair]
+
+            if self._book_buffer:
+                del self._book_buffer[std_pair]
 
         create_task(_concurrent_snapshot())
 


bmoscon:
@nirvana-msu does this fix your issue? Seems reasonable to me, but I haven't tested it

nirvana-msu:
@jinusean no it definitely does not fix it. It may make the exception go away, but the effect is even worse as it would instead lead to silently generating incorrect order book state. Actually, it doesn't even make exception go away - `if self._book_buffer` condition being `True` does not imply that there is `std_pair` in the dict, so `del self._book_buffer[std_pair]` will still often raise.

But that's not important.

The fundamental issue here is that due to feed restarts, there may be **multiple** of these `_concurrent_snapshot` coroutines **competing** with each other to do the same thing. There will never be any good from having multiple `_concurrent_snapshot` executing at the same time, regardless of how we change the code inside. The only way to guarantee correct order book state, is to ensure that a snapshot is overlayed with all subsequent order book updates in the correct order. When you have multiple of these coroutines running at the same time, they'll screw it up because the operations are non-atomic and their order it not guaranteed. For instance, coroutine `A` pops first update from `_book_buffer`, and coroutine `B` pops second update. It could be that `B` actually applies second udpate before `A` applies first (we cannot rely on the order in which coroutines will execute on the loop). Further, another one of these couroutines may at any point reset the snapshot - and if the snapshot it sets actually preceeded some of the updates that have already been applied (could easily be as the order of HTTP responses is not guaranteed to match order of requests), there'd be no way to recover correct state.

**TL;DR** is that there is no way for this to work correctly when there is a possibility of more than one `_concurrent_snapshot` executing concurrently. The correct way to fix it should be to cancel old tasks before starting new one on a feed restart - i.e. to guarantee that there will only be one `_concurrent_snapshot` executing for this pair at a time.

jinusean:
Ah understood. So a cache of the running tasks and a cancellation of those tasks should suffice?

nirvana-msu:
Something if this sort. You'll probably also need to await cancelled tasks to avoid messages from asyncio that an exception has never been retrieved. So possibly something along the lines of:

```python
    def _reset(self):
        self._l2_book = {}
        self.last_update_id = {}

        if self.concurrent_http:
            for task in self._concurrent_snapshot_task_cache:
                task.cancel()
                try:
                    await task
                except asyncio.CancelledError:
                    pass
            # buffer 'depthUpdate' book msgs until snapshot is fetched
            self._book_buffer: Dict[str, deque[Tuple[dict, str, float]]] = {}
```

But even then, I think there'd still be a possibility that a previous stream (the one currently running, preceding the reset) just didn't get to start the `_concurrent_snapshot` task yet (so its not yet in the cache) - but will do so soon, leading again to a race condition. 

i.e. the issue here is that we have two connections running / processing updates at the same time. Perhaps a better way to deal with it would be to ensure a connection, along with all pending tasks, is cancelled at at a higher level (in `connection_handler`?) before resetting and initiating a new one. Clearly that doesn't seem to happen at the moment (otherwise we'd never run into this race condition). I don't understand internals well enough yet to tell for sure - @bmoscon any advice?

bmoscon:
maybe just use an asyncio mutex/semaphore so only one task can run for a symbol at a time?

nirvana-msu:
But it’s not just a matter of only having one run at a time. We have to ensure a task from an older connection would never be run after a task from a newer connection. If it does, it can screw up order book state. We somehow need to ensure that an old connection is completely cancelled with all corresponding tasks, before we proceed with handling book state in a new connection.

bmoscon:
perhaps this should just be removed them, it seems like its adding a lot of unnecessary complexity

nirvana-msu:
What should be removed?

bmoscon:
the parallel snapshot code. Its kind of unnecessary. The user can get the same behavior by creating a feed per symbol

nirvana-msu:
So some way of doing this concurrently is definitely needed - it’s a great boost to performance. I’m not sure if a starting a feed per symbol if feasible when e.g. subscribing to each symbol on `Binance`? Are there rate limits on number of websocket connections?

Maybe the fix isn’t so hard.. I’m not sure if what I described above (with a possibility of a task not yet being in the cache when we reset, but added later) is actually possible - I’ll need to check what exactly happens when a connection is reset. Let me dig in the internals a bit to understand it better.

nirvana-msu:
Ok so looking at the code, I don't think the more complicated scenario I was describing is possible. By the time the new connection is created (and state is being reset), there should be nothing else running from the old connection, except for those dangling `_concurrent_snapshot` tasks.

So unless I'm mistaken, it should be sufficient to simply keep track of those running `_concurrent_snapshot` tasks in a cache, and cancel them (like I showed in https://github.com/bmoscon/cryptofeed/pull/673#issuecomment-932958547) when resetting.

@jinusean would you take a go at it?

jinusean:
@nirvana-msu I've added the recommended changes but I'm still getting a KeyError on the initial cancellation. Could you take a look?

villekuosmanen:
Seems like a tough concurrency bug, I was trying to think of good solutions but couldn't come up with any.

It feels like the biggest problem here is how `book_buffer` is shared between coroutines, even though it's used as a local store for buffered messages. If each coroutine had its own `book_buffer`, I don't think having multiple ones running at the same time would really matter since the sequence numbers would help us filter out duplicate updates. Of course, if you made `book_buffer` local to each coroutine, we would still need a way to append new book messages to them and I don't know if there's a neat way to do that in Python.

I'll definitely check if this bug exists in the refresh snapshot logic in #606, but since these snapshots are refreshed at most once per minute I think we should be fine on that front.

jinusean:
One thing I can’t figure out is why the snapshot coroutine continues to run even after it’s cancelled. On successive resets the keyerror is not raised.

nirvana-msu:
I can suggest one way how to tackle the second issue. We need to refactor the code to avoid ever having to cancel the task from within itself. It means we cannot simply call `self._reset()` within `_check_update_id` (because it could be called from inside `_concurrent_snapshot` task). But what we could do, is use exceptions for flow control instead.

So instead of calling `self._reset()` within `_check_update_id`, we raise a custom error like `InconsistentBookStateError`. Then we'd need to handle this error differently whether we catch it inside `_concurrent_snapshot`, or outside it. If caught outside  `_concurrent_snapshot` we can just `await self._reset()`, but if caught inside it, we need some special handling. One way I can think of would be to pop the task itself from the cache of tasks (as we want to avoid cancelling ourselves from inside), then  `await self._reset()` to ensure all other tasks are cancelled, and finally we end the task itself by just returning from it.

bmoscon:
@jinusean - will need to rebase with the latest changes, then I think the suggestion from @nirvana-msu can be implemented and we'll be good to go here

bmoscon:
@jinusean - I reverted most of the changes for concurrent snapshot, so once you have it fixed in your codebase you'll need to re-integrate based off of what I have. Too many bugs in the binance codebase, so I had to remove that large set of changes so I can push out a release while I wait for these issues to be resolved.

PR #717, opened 2021-11-21
Fix Bybit OrderInfo keys from msg/data

Acid3croco:
Update data extraction in _order() method to match right keys names

- [x] - Tested
- [x] - Changelog updated

### DIFF
--- a/cryptofeed/exchanges/bybit.py
+++ b/cryptofeed/exchanges/bybit.py
@@ -481,7 +481,7 @@ class Bybit(Feed):
                 Decimal(data['price']),
                 Decimal(data['cum_exec_qty']),
                 Decimal(data['qty']) - Decimal(data['cum_exec_qty']),
-                self.timestamp_normalize(data["update_time"]),
+                self.timestamp_normalize(data.get('update_time') or data.get('O') or data.get('timestamp')),
                 raw=data,
             )
             await self.callback(ORDER_INFO, oi, timestamp)
--- a/cryptofeed/exchanges/bybit.py
+++ b/cryptofeed/exchanges/bybit.py
@@ -475,14 +475,14 @@ class Bybit(Feed):
                 self.id,
                 self.exchange_symbol_to_std_symbol(data['symbol']),
                 data["order_id"],
-                BUY if msg["side"] == 'Buy' else SELL,
+                BUY if data["side"] == 'Buy' else SELL,
                 order_status[data["order_status"]],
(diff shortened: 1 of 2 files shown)


bmoscon (approved):

bmoscon (changes requested):
see comments inline

cryptofeed/exchanges/bybit.py
```diff
@@ -475,14 +475,14 @@ async def _order(self, msg: dict, timestamp: float):
...
                 order_status[data["order_status"]],
                 LIMIT if data['order_type'] == 'Limit' else MARKET,
                 Decimal(data['price']),
-                Decimal(data['cumQuantity']),
-                Decimal(data['qty']) - Decimal(data['cumQuantity']),
-                self.timestamp_normalize(data["updateTime"]),
-                raw=data
+                Decimal(data['cum_exec_qty']),
+                Decimal(data['qty']) - Decimal(data['cum_exec_qty']),
+                self.timestamp_normalize(data["update_time"]),
```
bmoscon, line 484: update_time is only available in USDT perpetuals, the others look like they support the `timestamp` field. Can you change this to use `update_time` if it exists in the json, otherwise use `timestamp`?

Acid3croco: Hey, I've added both timestamp and 'O' from spot market also

cryptofeed/exchanges/bybit.py
```diff
@@ -481,7 +481,7 @@ async def _order(self, msg: dict, timestamp: float):
                 Decimal(data['price']),
                 Decimal(data['cum_exec_qty']),
                 Decimal(data['qty']) - Decimal(data['cum_exec_qty']),
-                self.timestamp_normalize(data["update_time"]),
+                self.timestamp_normalize(data.get('update_time') or data.get('O') or data.get('timestamp')),
```
bmoscon, line 484: looks good now. What is the "O" field? I dont think I saw that in the docs?

bmoscon (approved):

cryptofeed/exchanges/bybit.py
bmoscon: ah ok, great!

cryptofeed/exchanges/bybit.py
Acid3croco: For the spot endpoint, they use O for timestamp I believe
https://bybit-exchange.github.io/docs/spot/#t-executionreport

PR #722, opened 2021-11-27
Fix auth issue

mariusba:

- [x] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass


bmoscon:
@mariusba - there isnt any description of this, nor have you indicated if this has even been tested or not. Please add more context / information to this so I can evaluate it

mariusba:
it is tested.
The issue here, that token refreshing does not work, as self.token = None is executed after self.Token is set by address() function

bmoscon (approved):

PR #734, opened 2021-12-04
Move data-processing from Mixin classes to the base RestExchange class (Binance demo).

D-Stacks:
Hello,

This code isn't fully fledged out, and probably shouldn't be merged for a while, if at all, but I have decided to open a draft pull none-the-less. it pertains to point no. 4 in #596. Mainly, I just want to gauge if there is any interest in my idea for an overhaul of the rest exchange classes, and possibly get some feedback, before I invest my time further.

Currently the pull draft consists of a small proof of concept, consisting of the BinanceRestMixin. 

The basic premise is to use keymaps to extract the required data from differing API responses across different exchanges. For this, a new `Keymap` class is implemented. In principle, this means standardized data responses can be extracted from any exchange, by simply defining a Keymap beforehand, and using the Keymap in just one `std_response = my_keymap_instance(response.text)` call. Currently, this call retrieves all the required data, but in an unformatted fashion. For this purpose post-processing functions are employed. I do believe that it would be possible for the keymap to do all formatting, but it would make the initialization of the Keymap very convoluted.

Dynamic manipulation of payloads can be done with a new `Payload` class. This is required for some pooling calls such as the`def _historical_trades()` function seen in most RestMixins. A way to translate between 'std' and 'exchange' timestamps would be another prerequisite to manipulate some payloads (and responses) dynamically. 

It also adds class level dictionaries into the Mixin to cut down on the need for parameter routing for things like the request method and API endpoints. 

Furthermore, some new API specific defines are used in order to keep the code neat and to standardize the response.

All this opens up the possibility to do the processing in a base `Rest` class (currently, I coded the functionality into the `RestExchange` class) by simply passing a Keymap and payload for most purposes.

Additional parameters may be used to handle exchange specific oddities. One such example can be seen with the `RestExchange()._rest_trades` function in this branch. It accepts a max_timeframe argument to resolve an issue with pooling for historical trades and the fact that binance does not allow the interval between start_time and end_time to be greater than 1 hour. These oddities may be one of the drawbacks, but I suspect these are either rare occurrences, or, implemented on enough exchanges to warrant an additional parameter. In essence, the idea is to deal with this through parameter twiddling. 

### Advantages:
- Cut down on code duplication
- Maintainablity
- Streamlined creation of RestMixin classes 
- Guarantees standardization, since all processing is done homogeneously.

### Disadvantages:
- Less performant
- May be somewhat unintuitive to work with Keymaps
- Difficulties for dealing with exchange specific oddities.

### TO DO: (if I were to continue):
- [ ] - Widen the proof of concept by implementing the revamped design on a few other exchanges
- [ ] - Add some stability, exception handling and Logging into `class RestExchange:` , `class Payload:` and `class Keymap:` 
- [ ] - Widen scope to private api calls:
     - [ ] - Add auth functions / or classes in `cryptofeed.auth
     - [ ] - Restructure symbol parsing somewhat, create an asset mapping function at exchange level
- [ ]  - Implement some form of rate-limit error handling and rate limit control into Mixin's

### Tests:
- [ ] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass


bmoscon:
seems useful, but I've noticed that many of the removed methods have not been replaced. I know this is a WIP / Draft, so wondering what (if anything) you want from me?

D-Stacks:
Well, the idea is to replace the methods eventually. I suppose, my question is simply: would my layout / design be implemented if everything were to run comprehensibly and smoothly (and all the old methods are back in place)? I can understand that introducing custom classes such as Keymaps and Payloads might make it all a bit unintuitive and there probably will be performance penalties (in case this is of any concern regarding rest calls).

bmoscon:
@D-Stacks - I think is a good idea. I'd definitely take the PR, but all the existing functionality would need to be replaced for binance before I'd accept it.

D-Stacks:
okay, I will find some time over christmas to implement authetification classes that can be called within the async connection, and restructure the symbol parsing (would also like to create an async Symbols.load_all(), while I am at it).

bmoscon:
since this hasnt seen any activity in over a month I'm going to close it. Feel free to re-open for file a new PR if you ever pick this back up.

PR #753, opened 2021-12-29
Fix funding rate zero

ablochs:
### When funding rate or predicted rate is 0, Funding.to_dict() returns None but should return 0. This causes errors, for example, when using FundingPostgres callback.

- [x] - Tested
- [ ] - Changelog updated
- [x] - Tests run and pass


bmoscon (approved):

bmoscon:
thanks for the PR!

PR #757, opened 2022-01-02
(bugfix): rate limit and fetch ending rest OKEx

bastienjalbert:
OKEx candles api endpoint are time reversed (start at end attribute and finish at start attribute). So the loop should decrease end attribute.
Requests must finished when len of data < 300 (since the current rate limit specified is 300)

- [X] - Tested
- [ ] - Changelog updated
- [X] - Tests run and pass

### DIFF
--- a/cryptofeed/exchanges/mixins/okex_rest.py
+++ b/cryptofeed/exchanges/mixins/okex_rest.py
@@ -52,11 +52,13 @@ class OKExRestMixin(RestExchange):
                            True, self._datetime_normalize(e[0]), raw=e) for e in data]
 
             yield data
-            if not end or len(data) < 10000:
+            if not end or len(data) < 300:
                 break
-            start = data[-1].start + offset
+            end = data[-1].stop - offset
 
-            await asyncio.sleep(1 / self.request_limit)
+            # TODO : optimize and use the 20 requests per 2 seconds slots
+            # TODO : Need to calculate current number of requests per 2 seconds slot and wait enough to start other requests
+            await asyncio.sleep(2)
 
     def _to_isoformat(self, timestamp):
         """Required for okex (ISO 8601)


cryptofeed/exchanges/mixins/okex_rest.py
```diff
@@ -52,11 +52,13 @@ async def candles(self, symbol: str, start=None, end=None, interval='1m', retry_
                            True, self._datetime_normalize(e[0]), raw=e) for e in data]
 
             yield data
-            if not end or len(data) < 10000:
+            if not end or len(data) < 300:
                 break
-            start = data[-1].start + offset
+            end = data[-1].stop - offset
 
-            await asyncio.sleep(1 / self.request_limit)
```
bmoscon, line 59: not sure why you would undo my correction here, but please change this back to what I have since 1) it uses asyncio and you cant put sleeps in async code (except in very specific situations) and 2) it correctly sleeps based on the exchange's rate limit

bmoscon (changes requested):

bmoscon (changes requested):
see previous comment

bastienjalbert:
> see previous comment

Not sure to understand. Did you see my comment ?

bmoscon:
no, I do not

bastienjalbert:
May you take a look ? I commented above your first comment.

bmoscon:
did you make a draft comment or something? I can post a screenshot, but there is no comment/reply to my comment in the code.

bastienjalbert:
The comment is "pending" : 
image

bmoscon:
yes, pending means you didn't post it. if its 20 per 2 seconds, then the limit (in okex.py under exchanges/) just needs to be changed to 10 (instead of 20).

cryptofeed/exchanges/mixins/okex_rest.py
bmoscon: if its 20 per 2 seconds, then the limit (in okex.py under exchanges/) just needs to be changed to 10 (instead of 20).

FPB-Indicators: I tested with 1/10 as sleeping time. It still reach rate limit.

bastienjalbert: As I said, I tried, without success (hit rate limit).

bmoscon: ok, seems like a bug or another issue. I dont think the solution is to change it to 1 every 2 seconds

bastienjalbert: I agree, but right now my code is not at all optimized but working. I'll optimize it in the near future.

bmoscon: probably should just change the limit in okex.py to 6, since the documentation indicates this is the default rate limit for the REST API. Some endpoints support other limits

bastienjalbert: I tried with 1/6, still the same. May I missed something ?

Testing class launch :
```python
import asyncio
from decimal import Decimal

from cryptofeed.exchanges import OKEx
from cryptofeed.types import Candle

o = OKEx()


def teardown_module(module):
    try:
        loop = asyncio.get_running_loop()
    except RuntimeError:
        loop = asyncio.new_event_loop()
    loop.run_until_complete(o.shutdown())


class TestOKExRest:

    async def test_candles(self):
        expected = Candle(
            o.id,
            'BTC-USDT',
            1609459200.0,
            1609459260.0,
            '1m',
            None,
            Decimal('28914.8'),
            Decimal('28959.1'),
            Decimal('28914.8'),
            Decimal('28959.1'),
            Decimal('13.22459039'),
            True,
            1609459200.0
        )
        ret = []
        async for data in o.candles('BTC-USDT', start='2021-01-01 00:00:00', end='2021-01-11 10:00:59', interval='1m'):
            print(data)
            ret.extend(data)

        assert len(ret) == 1
        assert ret[0] == expected

r = TestOKExRest()

loop = asyncio.get_event_loop()
forecast = loop.run_until_complete(r.test_candles())
loop.close()
```

The output :
```bash
[exchange: OKEX symbol: BTC-USDT start: 1610359200.0 ...
[exchange: OKEX symbol: BTC-USDT start: 1610341260.0 ...
[exchange: OKEX symbol: BTC-USDT start: 1610323320.0 ...
[exchange: OKEX symbol: BTC-USDT start: 1610305380.0 ...
OKEX.http.0: encountered a rate limit for address https://www.okex.com/api/spot/v3/instruments/BTC-USDT/history/candles?start=2021-01-10T14:04:00.000Z&end=2021-01-01T00:00:00.000Z&granularity=60&limit=300, retrying in 60 seconds
```

bmoscon (approved):

PR #806, opened 2022-03-08
fix: QuestDb backend column types

ideoma:
Fixes
- Makes received_timestamp_int column a TIMESTAMP
- Makes side, type column in trade SYMBOL
- Removes redundant timestamp= token

- [ ] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass

### DIFF
--- a/cryptofeed/backends/quest.py
+++ b/cryptofeed/backends/quest.py
@@ -74,10 +74,10 @@ class BookQuest(QuestCallback):
         super().__init__(*args, **kwargs)
         self.depth = depth
 
-    async def __call__(self, book, receipt_timestamp_int: float):
+    async def __call__(self, book, receipt_timestamp: float):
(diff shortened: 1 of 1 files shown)


cryptofeed/backends/quest.py
```diff
@@ -68,11 +74,12 @@ def __init__(self, *args, depth=10, **kwargs):
         super().__init__(*args, **kwargs)
         self.depth = depth
 
-    async def __call__(self, book, receipt_timestamp: float):
+    async def __call__(self, book, receipt_timestamp_int: float):
```
bmoscon, line 77: the value passed in here to receipt timestamp is definitely a float, so curious why you changed the name?

ideoma: typo, fixed

bmoscon (approved):

PR #820, opened 2022-04-10
fix: pass sent timestamp in binance snapshot

shaunc:

- [ ] - Tested
- [ ] - Changelog updated
- [ ] - Tests run and pass

Unfortunately, I had trouble getting the tests running locally. Also didn't know where exactly test should go. However, this fix is very small & localized.

### DIFF
--- a/cryptofeed/exchanges/binance.py
+++ b/cryptofeed/exchanges/binance.py
@@ -279,13 +279,19 @@ class Binance(Feed, BinanceRestMixin):
                     max_depth = d
                     break
 
+        req_time = time.time()
         resp = await self.http_conn.read(self.rest_endpoints[0].route('l2book', self.sandbox).format(pair, max_depth))
         resp = json.loads(resp, parse_float=Decimal)
+        if 'E' not in resp:
+            resp_time = time.time()
+            timestamp = req_time + (resp_time - req_time) / 2
+        else:
+            timestamp = self.timestamp_normalize(resp['E'])
 
         std_pair = self.exchange_symbol_to_std_symbol(pair)
         self.last_update_id[std_pair] = resp['lastUpdateId']
         self._l2_book[std_pair] = OrderBook(self.id, std_pair, max_depth=self.max_depth, bids={Decimal(u[0]): Decimal(u[1]) for u in resp['bids']}, asks={Decimal(u[0]): Decimal(u[1]) for u in resp['asks']})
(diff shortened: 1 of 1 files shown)


bmoscon:
You'll need to test that confirm that this doesnt break any of the other binance exchanges since they use a shared codebase

cryptofeed/exchanges/binance.py
```diff
@@ -285,7 +285,7 @@ async def _snapshot(self, pair: str) -> None:
         std_pair = self.exchange_symbol_to_std_symbol(pair)
         self.last_update_id[std_pair] = resp['lastUpdateId']
         self._l2_book[std_pair] = OrderBook(self.id, std_pair, max_depth=self.max_depth, bids={Decimal(u[0]): Decimal(u[1]) for u in resp['bids']}, asks={Decimal(u[0]): Decimal(u[1]) for u in resp['asks']})
-        await self.book_callback(L2_BOOK, self._l2_book[std_pair], time.time(), timestamp=None, raw=resp, sequence_number=self.last_update_id[std_pair])
+        await self.book_callback(L2_BOOK, self._l2_book[std_pair], time.time(), timestamp=self.timestamp_normalize(resp.get('E')), raw=resp, sequence_number=self.last_update_id[std_pair])
```
bmoscon, line 288: this fails for some variations of binance, so please fix and re-test and then I will merge

bmoscon (changes requested):

cryptofeed/exchanges/binance.py
shaunc: Right -- I can make it fail myself now ... will try to get a fix in later today. Thanks!

cryptofeed/exchanges/binance.py
```diff
@@ -279,13 +279,19 @@ async def _snapshot(self, pair: str) -> None:
                     max_depth = d
                     break
 
+        req_time = time.time()
         resp = await self.http_conn.read(self.rest_endpoints[0].route('l2book', self.sandbox).format(pair, max_depth))
         resp = json.loads(resp, parse_float=Decimal)
+        if 'E' not in resp:
+            resp_time = time.time()
+            timestamp = req_time + (resp_time - req_time) / 2
```
bmoscon, line 287: the timestamp field in the data object must always come from the exchange. you cannot generate a timestamp and put it in there. if the exchange does not provide a timestamp, it should be none, like the code currently is

bmoscon (changes requested):

cryptofeed/exchanges/binance.py
shaunc: Ok -- resubmitted.

Issue #828, opened 2022-04-19
Phemex Error


bazickoff:
**Describe the bug**

after running  examples/demo.py

I got this 
2022-04-19 15:09:41,272 : DEBUG : HTTPSync: requesting data from https://api.phemex.com/exchange/public/cfg/v2/products
2022-04-19 15:09:41,606 : DEBUG : PHEMEX.ws.1: connecting to wss://phemex.com/ws
2022-04-19 15:09:41,952 : DEBUG : PHEMEX.ws.1: Sending subscribe request to public channel: {'id': 1, 'method': 'orderbook.subscribe', 'params': [['uBTCUSD']]}
2022-04-19 15:09:42,024 : WARNING : PHEMEX.ws.1: Error from exchange {'error': {'code': 6001, 'message': 'invalid argument'}, 'id': 1, 'result': None}
[demo.log](https://github.com/bmoscon/cryptofeed/files/8513731/demo.log)

**To Reproduce**
run examples/demo.py

**Operating System:**
 - windows

**Cryptofeed Version**
- [v2.2.2]  installed pip install cryptofeed

trading8521:
got the same issue

The1iAmNeo:
@bazickoff can you tell me how you got this installed on windows?

PR #833, opened 2022-04-25
Fixed issue 828 (Phemex websocket subscription error)

sheungon:

- [x] - Tested
- [x] - Changelog updated
- [ ] - Tests run and pass

The original issue,
https://github.com/bmoscon/cryptofeed/issues/828

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,14 @@
 
 ### 2.2.3
  * Feature: Authenticated channel support for Bitget
+ * New Exchange: AscendEX Futures
+ * Update: AscendEX, add sandbox endpoint. Add channel filter.
+ * Update: Binance, add sandbox endpoint.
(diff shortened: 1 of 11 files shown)


sheungon:
> no, I'm not just changing the variable name from `symbol` to `symbols`
but also changed the variable handling from a single object to a list.
i.e. `[symbol]` <-- this make the list become a 2d list
to `symbols` <-- this use the variable as list directly
AND
`[symbol, self.candle_interval_map[self.candle_interval]]` <-- this, again the `symbol` list put in the list to become a list of list
to `[*symbols, self.candle_interval_map[self.candle_interval]]` <-- this appending every element in the list to the list

The reason of this bug was because `symbol` was a list. To avoid more confusion, it was renamed to `symbols`.

bmoscon:
oh I see, you added the star too, very subtle, sorry I missed it!

PR #853, opened 2022-05-23
Authenticated data channels support (Redis - streams & sorted sets)

AlgoQ:
Support for the authenticated data channels for Redis (Streams and Sorted Sets).

### DIFF
--- a/cryptofeed/backends/rabbitmq.py
+++ b/cryptofeed/backends/rabbitmq.py
@@ -7,6 +7,8 @@ associated with this software.
 import asyncio
 from collections import defaultdict
 
+from parso import ParserSyntaxError
+
 import aio_pika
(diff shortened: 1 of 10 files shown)


AlgoQ:
I will create another pull request for the other backends when I find the time (probably later this week).

cryptofeed/backends/rabbitmq.py
```diff
@@ -7,6 +7,8 @@
 import asyncio
 from collections import defaultdict
 
+from parso import ParserSyntaxError
```
bmoscon, line 10: this appears to not be used

bmoscon (changes requested):

bmoscon (approved):

Issue #516, opened 2021-06-12
Clickhouse support?


zoh:
Hello!

Who uses this ?  https://clickhouse.tech/

Need feeder to clickhouse.

bmoscon:
Seems like something that can be added to the backend

PR #898, opened 2022-08-28
Clickhouse Backend

EzePze:
Adds support for a ClickHouse backend. Based on the Postgres backend but using async ClickHouse drivers. Requested by #516 

- [x] - Tested: Demo file created and example tables provided -- almost one-to-one with the Postgres demo
- [x] - Changelog updated
- [x] - Tests run and pass


EzePze:
Tried to go that route initially but there's enough small changes between the two drivers that mean almost everything has to be rewritten anyway (e.g. formatting every new row as tuples instead of strings) -- though could def be worthwhile making a parent class for SQL-like DBs that share the same methods.

bmoscon:
old PR, closing

PR #973, opened 2023-05-26
deribit parse spot type

akasimo:
Deribit recently added spot type, which do not have the "settlement_period" key, leading to errors


bmoscon:
sorry, I didnt realize this had been opened and this has already been resolved

PR #999, opened 2023-11-21
Use hasattr() for aggregate callbacks

noahfrn:
`callable()` returns True for classes that override the `__call__` method, which means the existing logic doesn't properly handle all backends (using a BookPostgres callback as the handler will break, for example). 
This change uses `hasattr()` to explicitly determine if the given handler:
- Is a  class
- Has a start method
- Has a stop method

### DIFF
--- a/cryptofeed/backends/postgres.py
+++ b/cryptofeed/backends/postgres.py
@@ -8,6 +8,7 @@ from collections import defaultdict
 from datetime import datetime as dt
 from typing import Tuple
 
+import logging
 import asyncpg
 from yapic import json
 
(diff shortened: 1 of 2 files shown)


cryptofeed/backends/postgres.py
```diff
@@ -8,6 +8,7 @@
 from datetime import datetime as dt
 from typing import Tuple
 
+import logging
```
bmoscon, line 11: everything else is using the log.py file functionality, why did you opt to use the logging package here?

bmoscon (changes requested):

bmoscon:
feel free to re-open this if you have time to update the PR. closing for now

noahfrn:
Apologies for the delay on this, have re-opened #1013 with just the aggregate callback change.

PR #1067, opened 2024-11-17
fix: lock websockets version range

otomarukanta:

- [x] - Tested
- [x] - Changelog updated
- [ ] - Tests run and pass

The read_limit argument has been removed from websockets as of version 14.0. We have changed to using a version below 14.0.

https://github.com/python-websockets/websockets/blob/main/docs/howto/upgrade.rst#read_limit


bmoscon:
I resolved the issues so the latest version will work now

PR #1110, opened 2026-05-26
Fix QuestDB BookQuest IndexError when depth exceeds book levels (#1042)

### LINKED ISSUE #1042: quest backend receiving book with less depth that specified and crash
**Describe the bug**

**To Reproduce**
* Provide small code sample that reproduces the issue
* Provide any tracebacks, if applicable
it seems the reproduction depends on the live market condition, and not easily reproducible. The traceback is in the function of `BookQuest.__call__` where the message is assembled,  it reports either the `book.book.bids` or `book.book.asks` has no index 8 or 9(I set depth to 10). After I changed the `self.depth` to `len(book.book.bids)` and `len(book.book.asks)`, it does not report exception no more.

**Expected behavior**

**Screenshots**

**Operating System:**
 - macOS, linux, etc

**Cryptofeed Version**
- please list the version you are using, and how it was installed (from source, pypi, etc).
2.4.0

dimitryLMEX:
## Problem

`BookQuest.__call__` iterates `range(self.depth)` unconditionally for both bids and asks. When the order book has fewer levels than `self.depth`, calling `book.book.bids.index(i)` raises an `IndexError`.

Closes #1042.

## Fix

Clamp the iteration range using `min(self.depth, len(...))` for both sides before building the field list:

```python
bid_depth = min(self.depth, len(book.book.bids))
ask_depth = min(self.depth, len(book.book.asks))
```

This is a minimal, non-breaking change — when the book is fully populated the behaviour is identical to before.

## Testing

Verified with `flake8 cryptofeed/backends/quest.py` (no warnings).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cryptofeed/backends/quest.py
+++ b/cryptofeed/backends/quest.py
@@ -75,7 +75,11 @@ class BookQuest(QuestCallback):
         self.depth = depth
 
     async def __call__(self, book, receipt_timestamp: float):
-        vals = ','.join([f"bid_{i}_price={book.book.bids.index(i)[0]},bid_{i}_size={book.book.bids.index(i)[1]}" for i in range(self.depth)] + [f"ask_{i}_price={book.book.asks.index(i)[0]},ask_{i}_size={book.book.asks.index(i)[1]}" for i in range(self.depth)])
+        bid_depth = min(self.depth, len(book.book.bids))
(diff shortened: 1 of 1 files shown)


bmoscon:

PR #1116, opened 2026-06-18
Fix Coinbase public market data auth

### LINKED ISSUE #1114: Coinbase: API keys required for public market data (symbols() raises, subscribe() signs unconditionally)
**Version:** 2.4.1 (code unchanged on current master)

The Coinbase adapter requires `key_id`/`key_secret` even for public market data, although Coinbase made Advanced Trade market data public in 2024. Two places enforce this:

1. `Coinbase.symbols()` ([coinbase.py#L77-L82](https://github.com/bmoscon/cryptofeed/blob/v2.4.1/cryptofeed/exchanges/coinbase.py#L77-L82)) raises unconditionally without credentials:

```python
if 'coinbase' not in config or 'key_id' not in config['coinbase'] or 'key_secret' not in config['coinbase']:
    raise ValueError('You must provide key_id and key_secret in config to retrieve symbols from Coinbase.')
```

2. `subscribe()` ([coinbase.py#L173-L185](https://github.com/bmoscon/cryptofeed/blob/v2.4.1/cryptofeed/exchanges/coinbase.py#L173-L185)) calls `get_private_parameters(self.config, ...)` unconditionally, which reads `config["coinbase"]["key_secret"]` and signs every subscription — so subscribing without keys fails even though the channels don't need a signature.

### What Coinbase actually requires now

- **REST:** `GET https://api.coinbase.com/api/v3/brokerage/market/products` is the public, unauthenticated twin of the authed `/api/v3/brokerage/products` the adapter currently uses for the symbol map. Verify with:

  ```
  curl -s 'https://api.coinbase.com/api/v3/brokerage/market/products' | head -c 300
  ```

  It returns the full product list with the same shape (`products`, `product_id`, `base_currency_id`, `quote_currency_id`, `quote_increment`), so `_parse_symbol_data` works as-is.

- **WebSocket:** `wss://advanced-trade-ws.coinbase.com` accepts unsigned subscriptions to the public market data channels (`market_trades`, `level2`, `heartbeats`). An unsigned message of the shape the adapter already builds:

  ```json
  {"type": "subscribe", "product_ids": ["BTC-USD"], "channel": "market_trades"}
  ```

  is accepted and streams trades (verified live).

Docs: https://docs.cdp.coinbase.com/advanced-trade/docs/ws-overview (public market data does not require authentication).

### Suggested fix

- Point the symbol-map REST call at `/market/products` (no headers needed) and drop the credential check in `symbols()`.
- In `subscribe()`, only merge in `get_private_parameters(...)` when credentials are actually configured, so authenticated use keeps working but TRADES/L2_BOOK run with zero config.

I have a working subclass doing exactly this and can open a PR if that's welcome.

DeepCogNeural:
## Summary
- load Coinbase symbols from the public `/market/products` endpoint so public symbol lookup no longer requires Coinbase API keys
- avoid adding Coinbase websocket signing fields when no API keys are configured, while preserving the signed branch when keys are present
- subscribe to and ignore the documented public `heartbeats` channel

Fixes #1114.

## Tests
- `.venv313/bin/python -m pytest tests/unit/test_coinbase.py -q` (`4 passed`)
- `git diff --check`

## Manual checks
- reproduced the pre-fix no-key `Coinbase.symbols(config={}, refresh=True)` failure as a credential `ValueError`
- reproduced the pre-fix no-key `subscribe()` path entering Coinbase signing and failing on missing `key_secret`
- verified with a mocked public products response that `Coinbase.symbols(config={}, refresh=True)` reads `/api/v3/brokerage/market/products` with `headers=None`
- verified with a recording connection that no-key public subscriptions send unsigned `market_trades` and `heartbeats` payloads
- verified with a keyed config that subscription payloads still include `api_key`, `timestamp`, and `signature`

### DIFF
--- a/cryptofeed/exchanges/coinbase.py
+++ b/cryptofeed/exchanges/coinbase.py
@@ -74,12 +74,12 @@ class Coinbase(Feed, CoinbaseRestMixin):
         return ret, info
 
     @classmethod
+    def _symbol_endpoint_prepare(cls, ep: RestEndpoint):
+        return f"{ep.address}/market/products"
+
+    @classmethod
     def symbols(cls, config: dict = None, refresh=False) -> list:
-        config = Config(config)
-        if 'coinbase' not in config or 'key_id' not in config['coinbase'] or 'key_secret' not in config['coinbase']:
-            raise ValueError('You must provide key_id and key_secret in config to retrieve symbols from Coinbase.')
-        headers = get_private_parameters(config, rest_api=True, endpoint='products')
-        return list(cls.symbol_mapping(refresh=refresh, headers=headers).keys())
+        return list(cls.symbol_mapping(refresh=refresh).keys())
 
     def __init__(self, callbacks=None, **kwargs):
         super().__init__(callbacks=callbacks, **kwargs)
@@ -179,7 +179,7 @@ class Coinbase(Feed, CoinbaseRestMixin):
                       "product_ids": product_ids,
                       "channel": chan
                       }
-            private_params = get_private_parameters(self.config, chan, product_ids)
+            private_params = get_private_parameters(self.config, chan, product_ids) if self.key_id and self.key_secret else {}
             if private_params:
                 params = {**params, **private_params}
             await conn.write(json.dumps(params))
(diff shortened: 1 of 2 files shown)


bmoscon:

Issue #1082, opened 2025-05-07
Order book sync logic vulnerable to edge cases


Joshuaalbert:
I will use Bitstamp L2 as an example, but I think the same flow is used in others.

Currently, you
1. subscribe, 
2. wait a few seconds, 
3. fetch a snapshot, 
4. and then drop all messages older than the snapshot. 

There are a few problems with this:
1. After step 1, you immediately start handling deltas, applying them to an empty L2 book, and initiating callbacks. These delta's mean nothing unless applied correctly to a snapshot. The callbacks will be triggered with garbage order books and the user has no way on knowing when/if they stop being garbage.
2. A delta can arrive while fetching a snapshot that should be applied AFTER the snapshot. But your code would reset the order book and those deltas would never be applied, leading to an inconsistent state.

The only perfectly correct way is as follows:
1. collect at least one delta.
2. if there is no snapshot yet, then fetch in a loop until snapshot['timestamp'] > delta['timestamp']
3. drop all delta['timestamp'] <= snapshot['timestamp']

In pseudo-code
```python
book = None
subscribe()
while True:
  delta = recv()
  if book is None:
    book = fetch_book(minimum_timestamp=delta['timestamp'])
  if delta['timestamp'] <= book['timestamp']:
    continue
  book += delta
```

This approach is 100% sure to be consistent regardless of straggler deltas, but requires at least one delta to trigger fetching the book. Low-volume markets don't always have order activity which is a problem. Thus, a second way is the following:

1. subscribe
2. put all deltas onto a queue
3. wait fetch wait
4. start handling queue

In pseudo-code
```python
q=queue()
def task_consume():
  subscribe()
  while True:
    delta = recv()
    q.put(delta)

def task_apply():
  while book is None:
    sleep(1)
  while True:
    delta = q.get()
    if delta['timestamp'] <= book['timestamp']:
      continue
    book += delta

create_task(task_consume())
create_task(task_apply())

sleep(safety) # allow any deltas to arrive before fetch
temp_book = fetch_book()
sleep(safety) # allow any stragglers to arrive after fetch
book = temp_book
```

This approach queues up the deltas and waits for some safety period before and after fetching the book, before applying the deltas. This is different from your current approaches which does not queue deltas initially. It's virtually perfect, because there are physical limits to straggler delay.

PR #1126, opened 2026-08-20
Fix Bitstamp L2 snapshot synchronization precision

mameikagou:

Bitstamp's REST order-book snapshot and websocket deltas both expose a `microtimestamp`, but the snapshot watermark was stored and compared at whole-second precision. That made updates within the snapshot's second impossible to order reliably: an update already represented by the snapshot could be applied again.

This keeps the existing synchronization flow and makes its boundary precise:

- store the REST snapshot watermark as an integer microtimestamp;
- discard websocket deltas at or before that watermark;
- apply the first delta after it;
- cover the before/equal/after cases with a deterministic unit test and update the Bitstamp playback expectation.

The connection handler already waits for `subscribe()` (including `_snapshot()`) before starting the websocket message handler, so this does not add a second queue or change connection ordering.

Related to #1082

### Verification

- `python -m pytest -q -s tests/unit/test_bitstamp.py 'tests/unit/test_exchange.py::test_exchange_playback[BITSTAMP]'` — 2 passed
- `ruff check cryptofeed/exchanges/bitstamp.py tests/unit/test_bitstamp.py tests/unit/test_exchange.py`
- `flake8 --ignore=E501 cryptofeed/exchanges/bitstamp.py tests/unit/test_bitstamp.py tests/unit/test_exchange.py`
- `git diff --check`

### Checklist

- [x] Tested
- [x] Changelog updated
- [x] Tests run and pass
- [x] Flake8 run and all relevant errors/warnings resolved
- [ ] Contributors file updated (optional)


bmoscon:

